PR #59, opened 2025-04-25
feat: add endpoint for ingesting webhooks forwarded from Sentry

joseph-sentry:
depends on https://github.com/codecov/umbrella/pull/392

add dictionary for routing request to a custom Sentry webhook handler
or the regular github handlers. this is because we don't want to process
all webhooks sent to us by sentry, and we also need custom processing
for other events, such as "pull_request"

use jwt auth permission class, since we'll be receiving these requests
via the codecov client in Sentry

emit logs and sentry exceptions on any exception during handling for
observability


apps/codecov-api/webhook_handlers/views/sentry.py
```diff
@@ -0,0 +1,84 @@
...
+    permission_classes = [JWTAuthenticationPermission]
+
+    def post(self, request):
+        with transaction.atomic():
+            # this looks weird but we're basically doing "dry runs"
+            # of the webhook handling
+
+            # this will eventually be removed once we validate it's working
+            # correctly in prod
+            transaction.set_rollback(True)
```
Swatinem, line 29: lets hope the assumption that we can use transaction rollback for this actually works out, lol

PR #60, opened 2025-04-28
Turn `labelanalysis` and `staticanalysis` views into noops

Swatinem:
(migrated from https://github.com/codecov/codecov-api/pull/1233)

As part of deprecating ATS, this turns all the views into noops that return empty objects.

Checking in with the requests that the CLI sends, returning an empty `external_id` and `absent_labels` for the labelanalysis will result in the CLI falling back to using the clientside requested labels.

Similarly, returning an empty list of `filepaths` means that the CLI will not do any further uploads.


joseph-sentry (approved):
just leaving a note here for anyone that comes back to look at this code in the future that the reason we have to create `staticanalyses/<uuid:external_id>/finish` is because the `action` decorator that previously existed on the `finish` method off `StaticAnalysisSuiteViewSet` was implicitly creating that route

PR #63, opened 2025-04-28
Clean up some `report_code` usage

Swatinem:
This removes some `report_code` usage where it is not being used.

It also simplifies the `disptach_upload_task` fn a bit, which I just couldn’t resist :-D

### DIFF
--- a/apps/codecov-api/upload/helpers.py
+++ b/apps/codecov-api/upload/helpers.py
@@ -714,49 +714,35 @@ def parse_headers(
 
 
 def dispatch_upload_task(
-    task_arguments: Dict[str, Any],
-    repository: Repository,
     redis: Redis,
-    report_type: Optional[CommitReport.ReportType] = CommitReport.ReportType.COVERAGE,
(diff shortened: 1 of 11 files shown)


apps/codecov-api/upload/helpers.py
```diff
@@ -714,49 +714,35 @@ def parse_headers(
...
-    redis.setex(
-        latest_upload_key,
-        3600,
-        timezone.now().timestamp(),
-    )
-    commitid = task_arguments.get("commit")
+        repo_queue_key = f"uploads/{repoid}/{commitid}/{report_type}"
+        latest_upload_key = f"latest_upload/{repoid}/{commitid}/{report_type}"
+
+    with redis.pipeline() as pipeline:
```
giovanni-guidini, line 734: I didn't know about this, it's nice.
Will it make all commands in the same connection (so it's faster) too?

giovanni-guidini (approved):
LGTM

PR #64, opened 2025-04-28
WIP: Move to a `uv` workspace

Swatinem:
This is a post-absorb rebase of #18


matt-codecov (commented):
could you preserve our current "requirements image" / "app image" distinction so we can cache the reqs image and cut down on build/CI time?

i actually already moved us to a unified requirements image that i think should work fine with a unified `pyproject.toml`. it has an `APP_DIR` argument which is expected to contain a `pyproject.toml` and `uv.lock` that deps will be installed from. currently `Makefile` passes [`apps/codecov-api` for api](https://github.com/codecov/umbrella/blob/ab0a1ac848acfa38f37de5772b722f06187d4990/Makefile#L39) and `apps/worker` for worker](https://github.com/codecov/umbrella/blob/ab0a1ac848acfa38f37de5772b722f06187d4990/Makefile#L72), but i think if you make both pass `.` it should... just work?

admittedly there would be a couple more parts to it:
- as you've observed, we currently don't really use uv. we just convert back to reqs.txt. but it looks like you know the appropriate way to install the uv env, so changing that shouldn't be major
- the reqs image cache key in `_self-hosted.yml` and `_build-app.yml` would need to be updated to no longer insert `output_directory` before `uv.lock` when hashing

i think that would do it for the requirements image, and then you would just need your changes to accept a requirements image as an argument the same way the existing `Dockerfile`s do.

(then to use a unified _app_ image you would want to get rid of the `output_directory` argument in `_build-app.yml` and `_self-hosted.yml` entirely and also copy the `tag.*`, `save.*`, `load.*`, and `push.*` targets from worker/api makefiles to the root makefile)

(sorry the existing setup is messy - i sort of anticipated i would be the one cleaning it all up so i did not comment very well or share a lot with the team about how it works)

PR #66, opened 2025-04-28
Use only one `ArchiveService`

Swatinem:
This switches over `worker` to the `archive_service` defined in `shared`.

Additionally, this also changes the `BundleAnalysisReportLoader` to use the `ArchiveService` constructor, instead of rolling its own.


giovanni-guidini (approved):
LGTM... both files seemed to export the same functions. But this is a lot of changes, let's keep an eye on the rollout 🙏

PR #68, opened 2025-04-28
Switch some `transactional_db` fixtures to `db`

Swatinem:
Similar to the case with `TransactionTestCase`, the `transactional_db` might be a bit slower, as is has to do more heavy lifting to setup/teardown the test. So lets switch to regular testcases, hopefully speeding up the tests slightly.

### DIFF
--- a/apps/worker/services/test_analytics/ta_process_flakes.py
+++ b/apps/worker/services/test_analytics/ta_process_flakes.py
@@ -1,7 +1,6 @@
 import logging
 from datetime import datetime
 
-from django.db import transaction
 from django.db.models import Q, QuerySet
 from redis.exceptions import LockError
 from shared.django_apps.reports.models import CommitReport, ReportSession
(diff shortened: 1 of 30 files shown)


apps/worker/services/test_analytics/ta_process_flakes.py
```diff
@@ -104,8 +103,6 @@ def process_flakes_for_commit(repo_id: int, commit_id: str):
         update_fields=["end_date", "count", "recent_passes_count", "fail_count"],
     )
 
-    transaction.commit()
```
matt-codecov, line 107: was this forced by the removal of `transactional_db`? i think it's safe because we turned Django autocommit back on, but this is much more significant than the rest of the change

matt-codecov (approved):
gave it a skim with a slightly closer look at the removed commits and changed test

apps/worker/services/test_analytics/ta_process_flakes.py
Swatinem: Yes, otherwise the tests would fail because they were not running in a transaction.
And yes, autocommit should make this unnecessary.

PR #72, opened 2025-04-28
build: run TA timeseries migrations

joseph-sentry:
we need to:
- create the TA db in timescale
- run the migrate_timeseries.py script to apply migrations

### DIFF
--- a/tools/devenv/scripts/start-worker.sh
+++ b/tools/devenv/scripts/start-worker.sh
@@ -21,6 +21,7 @@ fi
 
 python manage.py migrate
 python manage.py migrate --database "timeseries"
+python manage.py migrate_timeseries.py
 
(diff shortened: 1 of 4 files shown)


michelletran-codecov (approved):

Swatinem (approved):
nice. I guess this will solve all the TA tests currently failing locally?

tools/devenv/scripts/start-worker.sh
```diff
@@ -21,6 +21,7 @@ fi
 
 python manage.py migrate
 python manage.py migrate --database "timeseries"
+python manage.py migrate_timeseries.py
```
matt-codecov, line 24: this is broken, `migrate_timeseries.py` is not a `manage.py` command

PR #73, opened 2025-04-28
docker: build with host network

joseph-sentry:
I've been having problems building the requirements images since fetching the debian or cargo reqs will randomly fail, building with the host network fixes that


Swatinem (approved):
sure, why not. though I don’t know enough about docker to really know why this would be needed.

joseph-sentry:
> sure, why not. though I don’t know enough about docker to really know why this would be needed.

idk the mechanics behind it but sometimes my requirements image build just fails and this seems to fix it 🤷

PR #75, opened 2025-04-28
fix: add shared test junit/coverage artifacts to gitignore

matt-codecov:
it would make sense to merge all three projects' `.gitignore` into a top-level `umbrella` one but that's a different commit


ajay-sentry (approved):
makes sense, this is for cases where we rename / prefix our junit and coverage files?

PR #81, opened 2025-04-29
Fully remove and disallow non-default `report_code`

Swatinem:
This removes all the places where a non-default `report_code` is being passed around and used. It also changes the view serializer to reject any non-"default" `code` as well.

### DIFF
--- a/apps/codecov-api/upload/views/base.py
+++ b/apps/codecov-api/upload/views/base.py
@@ -68,9 +68,10 @@ class GetterMixin(ShelterMixin):
         ] = CommitReport.ReportType.COVERAGE,
     ) -> CommitReport:
         report_code = self.kwargs.get("report_code")
-        if report_code == "default":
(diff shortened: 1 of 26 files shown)


apps/codecov-api/upload/views/base.py
```diff
@@ -68,9 +68,10 @@ def get_report(
         ] = CommitReport.ReportType.COVERAGE,
     ) -> CommitReport:
         report_code = self.kwargs.get("report_code")
-        if report_code == "default":
-            report_code = None
-        queryset = CommitReport.objects.filter(code=report_code, commit=commit)
+        if report_code not in (None, "default"):
+            raise ValidationError("Report not found")
```
joseph-sentry, line 72: I would prefer this error to be more specific, mentioning that it was because a report code was specified

joseph-sentry (approved):

PR #84, opened 2025-04-29
Remove unused `BackfillTestInstances` and `UpdateBranch` tasks

Swatinem:
Found these while checking our own coverage. They don’t have tests, don’t seem to be invoked anywhere, and the `celery_config` also had a bug where both were assigned the same name, lol.


joseph-sentry (approved):
yeah these were one off tasks meant to be run manually, they're no longer needed!

PR #85, opened 2025-04-29
remove deprecated integration stuff, simplify github install webhooks

nora-codecov:
removing the deprecated integration flow means these two webhook events could be simplified.

modified them to share a `handle` function

### DIFF
--- a/apps/codecov-api/webhook_handlers/tests/test_github.py
+++ b/apps/codecov-api/webhook_handlers/tests/test_github.py
@@ -849,12 +849,13 @@ class GithubWebhookHandlerTests(APITestCase):
         owner.refresh_from_db()
         repo1.refresh_from_db()
         repo2.refresh_from_db()
-        # we no longer update these fields on the Owner or Repos
-        assert owner.integration_id == 12
-        assert repo1.using_integration == True
-        assert repo2.using_integration == True
-        assert repo1.bot is not None
-        assert repo2.bot is not None
+
+        assert owner.integration_id is None
+        assert repo1.using_integration == False
+        assert repo2.using_integration == False
+
+        assert repo1.bot is None
+        assert repo2.bot is None
 
         assert not owner.github_app_installations.exists()
 
@@ -973,7 +974,7 @@ class GithubWebhookHandlerTests(APITestCase):
         using_integration,
         repos_affected: None,
     )
-    def test_installation_with_other_actions(
+    def test_installation_with_other_actions_sets_owner_integration_id_if_none(
         self,
     ):
         installation_id = 44
@@ -996,6 +997,9 @@ class GithubWebhookHandlerTests(APITestCase):
                 "sender": {"type": "User"},
             },
         )
+
+        owner.refresh_from_db()
+        assert owner.integration_id is None  # no longer set this during install
 
         ghapp_installations_set = GithubAppInstallation.objects.filter(
             owner_id=owner.ownerid
@@ -1022,7 +1026,7 @@ class GithubWebhookHandlerTests(APITestCase):
         self,
     ):
         installation_id = 44
-        owner = OwnerFactory(service=Service.GITHUB.value)
+        owner = OwnerFactory(service=Service.GITHUB.value, integration_id=None)
 
         self._post_event_data(
             event=GitHubWebhookEvents.INSTALLATION_REPOSITORIES,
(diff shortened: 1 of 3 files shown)


apps/codecov-api/webhook_handlers/views/github.py
```diff
@@ -510,65 +470,70 @@ def _handle_installation_events(
...
+                    repositories_removed_service_ids = {
+                        obj["id"]
+                        for obj in request.data.get("repositories_removed", [])
+                    }
+                    repo_list_to_save = current_repos.union(
+                        repositories_added_service_ids
+                    ).difference(repositories_removed_service_ids)
+                    ghapp_installation.repository_service_ids = list(repo_list_to_save)
 
                 if action in ["suspend", "unsuspend"]:
```
matt-codecov, line 538: should this be de-indented by one? is suspend/unsuspend only relevant when `all_repos_selected` is `False`?

matt-codecov (approved):
accept with an important question. might be the cause of the test failures

apps/codecov-api/webhook_handlers/views/github.py
nora-codecov: ty for catching!! fixed!

apps/codecov-api/webhook_handlers/views/github.py
```diff
@@ -501,96 +461,95 @@ def _handle_installation_events(
...
-                    owner.github_app_installations.filter(
-                        installation_id=installation_id
-                    ).first()
-                )
-                if ghapp_installation is not None:
-                    ghapp_installation.delete()
-            # Deprecated flow - BEGIN
-            owner.integration_id = None
-            owner.save()
-            owner.repository_set.all().update(using_integration=False, bot=None)
```
nora-codecov, line 516: @giovanni-guidini wanted to check that we want to remove these steps, including the updates to repos

apps/codecov-api/webhook_handlers/tests/test_github.py
```diff
@@ -858,13 +849,12 @@ def test_installation_with_deleted_action_nulls_values(self):
...
-        assert repo2.using_integration == False
-
-        assert repo1.bot is None
-        assert repo2.bot is None
+        # we no longer update these fields on the Owner or Repos
+        assert owner.integration_id == 12
+        assert repo1.using_integration == True
+        assert repo2.using_integration == True
+        assert repo1.bot is not None
+        assert repo2.bot is not None
```
nora-codecov, line 857: @giovanni-guidini this is the consequence of removing that deprecated flow in the delete block - seems weird

apps/codecov-api/webhook_handlers/tests/test_github.py
```diff
@@ -960,10 +949,12 @@ def test_installation_repositories_update_existing_ghapp_all_repos(self):
...
                     "app_id": 15,
                 },
-                "repositories_added": [{"id": repo2.service_id}],
+                "repositories_added": [
+                    {"id": repo2.service_id, "node_id": "R_xCATxCAT"}
+                ],
                 "repositories_removed": [],
                 "repository_selection": "all",
-                "action": "deleted",
+                "action": "added",
```
nora-codecov, line 957: these fields were invalid for the `event` type, fixed to valid format/values
https://docs.github.com/en/webhooks/webhook-events-and-payloads#installation_repositories

apps/codecov-api/webhook_handlers/views/github.py
giovanni-guidini: wow 2 weeks, my bad... I must have missed this notification. I'm sorry.

I think if we want to remove these steps we also need to remove [the usage](https://github.com/codecov/umbrella/blob/a05cb0cfa1cce54f4ca6e3ecf5eda4b22e5d4b27/libs/shared/shared/bots/github_apps.py#L338) when getting bots. In theory we have moved the "management" of the github app integration fully to `GithubAppInstallation` objects, so that the `owner_integration_id` and `repo.using_integration` become irrelevant and can be removed from the table.

About bot being None... I'd keep that operation maybe. The consequence of removing the installation and not clearing up the bot - if an old bot exists - is that the notifications would be sent by that bot account. We had this happen to us at some point in time when notifications were sent by Eli's account (the bet setup)

So I'd keep that one operation in there.

apps/codecov-api/webhook_handlers/tests/test_github.py
giovanni-guidini: Well yeah... but these fields then basically become irrelevant (assuming you remove usage of them as well)

PR #86, opened 2025-04-29
move lint commands to root

nora-codecov:
I like running the lint commands from the root of the project I'm working on.

We had the lint commands duplicated across the 3 projects, this pr adds them to umbrella's Makefile and removes the duplicated code from the other Makefiles


matt-codecov (changes requested):
req changes since you at least have to un-delete the lint commands from the original makefiles to unbreak CI right now

PR #92, opened 2025-04-30
Move `ruff` linting to the toplevel workspace

Swatinem:
This creates a toplevel `pyproject.toml` that currently contains some devtools like `pre-commit`, `ruff` and `mypy`.

It also consolidates the various configuration files for these tools (except `mypy`), moving them to the toplevel, and removing them from each subproject. The same is also done for the `.vscode` config, which enables auto-formatting.

As another change, this also removes the per-subproject CI lint job in favor of just a single job, linting the whole repo at once.

---

Some of the commits are just auto-fixes, with the `C408` one being the largest, and probably one that could be reverted as the churn might be too much all at once.


.pre-commit-config.yaml
```diff
@@ -1,16 +1,7 @@
 repos:
-- repo: local
+- repo: https://github.com/astral-sh/ruff-pre-commit
```
ajay-sentry, line 2: this is sweet, love it

.pre-commit-config.yaml
```diff
@@ -1,16 +1,7 @@
...
+  rev: v0.11.7
```
ajay-sentry, line 3: is it possible to pin this to latest tho? I'm good either way tbh

Swatinem: that is the latest release as of now: https://github.com/astral-sh/ruff/releases
I don’t think you can actually pin to `latest` (as in the tag)
I also don’t think that would be a good idea, as that would just cause some confusion when some defaults are changing suddenly.

ajay-sentry (approved):
send it, one thought on the latest thing

matt-codecov (approved):

PR #94, opened 2025-04-30
add flare_cleanup to regular_cleanup

nora-codecov:
reorganizing flare cleanup task, adding it to nightly cleanup. The goal is for it to not break any other cleanup operations if it crashes, any errors it encounters will show up in sentry, and to be traceable with logs.

Last time we ran this manually it errored out with `"message": "Connection pool is full, discarding connection: [storage.googleapis.com](http://storage.googleapis.com/). Connection pool size: 10"`, so none of the files in the archive were deleted but the database fields were cleared.

Added its own method (inspired by `cleanup_files_batched`) where the archive path field (`_flare_storage_path`) is NOT cleared if the attempt to delete returns an error.

moved the `limit` down to `1000`, can increase when initial runs are successful.

https://github.com/codecov/internal-issues/issues/1029


apps/worker/services/cleanup/regular.py
```diff
@@ -1,15 +1,121 @@
...
+    while start < limit:
+        stop = start + batch_size if start + batch_size < limit else limit
+        batch = non_open_pulls_with_flare_in_archive.values_list("id", flat=True)[
+            start:stop
+        ]
+        if not batch:
+            break
+
+        # Get ids and paths together to track which were successfully deleted
+        id_path_pairs = list(
```
Swatinem, line 79: I think this is weird to select *just* the `id` first, and then select another field based on the first query. You can just select both fields without the `flat=True`, and use the slice for that single query.

apps/worker/services/cleanup/regular.py
```diff
@@ -1,15 +1,121 @@
...
+
+    # Process in batches
+    total_db_updated = 0
+    start = 0
+    while start < limit:
+        stop = start + batch_size if start + batch_size < limit else limit
+        batch = non_open_pulls_with_flare_in_db.values_list("id", flat=True)[start:stop]
+        if not batch:
+            break
+        n_updated = non_open_pulls_with_flare_in_db.filter(id__in=batch).update(
```
Swatinem, line 53: splitting into two queries does make more sense here, as it constrains the `update`. Which I don’t think you necessarily need to do, as that query might be trivial/fast enough to not matter.

apps/worker/services/cleanup/regular.py
```diff
@@ -31,6 +137,11 @@ def run_regular_cleanup() -> CleanupSummary:
         summary = cleanup_old_uploads(context)
         complete_summary.add(summary)
 
+        # Run flare cleanup as part of regular cleanup
+        # reduce limit for initial runs
+        flare_summary = cleanup_flares(context, limit=1000)
+        complete_summary.add(flare_summary)
```
Swatinem, line 143: Please note that the current cleanup task runs the queryset-based cleanup, and `Upload` cleanup first. The second will definitely time out, and then be resumed with a retry.
This is fully intentional, and there is nothing wrong with it. It just means that we will currently not be hitting this code in production.

I have plans to better "randomize" this, so that each individual cleanup can make progress periodically without being blocked on the others.
This is not a blocker for this PR though, which can land before that.

Swatinem (approved):

apps/worker/services/cleanup/regular.py
nora-codecov: oh interesting - I didn't realize that was the case. Could I move this task up so it runs before upload cleanup, so we could see if it even can run successfully?

apps/worker/services/cleanup/regular.py
nora-codecov: haha good call, fixing

apps/worker/services/cleanup/regular.py
Swatinem: https://github.com/codecov/umbrella/pull/112 is changing the way these multiple cleanup jobs are being scheduled, and I added a `TODO` directly to indicate where to put the flare cleanup for this to be picked up. There is still some randomization happening though.

PR #99, opened 2025-05-01
chore: suppress Sentry error from token refresh failure

spalmurray:
Prevent the error log that occurs when token refresh fails from bubbling up to Sentry.

Resolves https://codecov.sentry.io/issues/4530397997/events/ab8faf32f54247a9a3c9fac416a7a757/


ajay-sentry (approved):
send it

PR #102, opened 2025-05-01
fix: add LC_ALL=C to makefile sort usage to normalize order

matt-codecov:
what an irritating bug

with this fixed, your mac should come up with the same requirements image tags as CI. with this + https://github.com/codecov/umbrella/pull/101 you should be able to pull prebuilt recs images and speed up your local dev experience


Swatinem (approved):
lol, I hate how the locale messes with such foundational tools :-(

PR #103, opened 2025-05-01
fix: make test flag bridge id bigint

joseph-sentry:


matt-codecov (approved):

nora-codecov (approved):

Swatinem (approved):
lgtm, though keep in mind that this will migration will rewrite the whole table, and the table is huge, and has caused us slowdowns in other dependent queries already.

So executing this migration (marked as risky) will probably cause a downtime.

PR #105, opened 2025-05-02
feat(ta-gql): handle receiving all branches filter

joseph-sentry:
add all branches special case for TA results branch filter in new TA
functionality


apps/codecov-api/graphql_api/tests/test_test_analytics.py
```diff
@@ -871,3 +868,68 @@ def test_gql_query_with_new_ta_v1(self, mocker, repository, snapshot):
...
+        # set the feature flag
+        mocker.patch("rollouts.READ_NEW_TA.check_value", return_value=True)
+
+        # read file from samples
+        storage = get_appropriate_storage_service()
+        try:
+            storage.create_root_storage(settings.GCS_BUCKET_NAME)
+        except BucketAlreadyExistsError:
+            pass
+        storage.write_file(
```
Swatinem, line 882: Why not use `mock_storage` for this? I believe when that is missing, this is targetting the minio running in docker?
Using the `mock_storage` would make things easier as you won’t have to create the bucket or clean up the file (which btw won’t run if the assertions fail, because there is no try/finally)

Swatinem (approved):

PR #110, opened 2025-05-02
fix: typo in lint CI name and unnecessary docker-compose dep

matt-codecov:
required checks are based on the name so before i make the lint check required i should fix the typo

also i originally made `worker` depend on `api` in `docker-compose.yml` so that the latter would run migrations first. turns out that didn't work and they still raced, so now there is `make devenv.migrate`. this PR removes that dependency which will slightly improve startup and teardown time

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -71,8 +71,6 @@ services:
       - postgres
       - redis
       - timescale
-      # Make worker block on API so API can run migrations first
-      - api
 
   # Dummy service to run shared tests
   shared:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -7,7 +7,7 @@ services:
       - ./tools/devenv/config:/config
     ports:
       - "${CODECOV_PORT-8080}:8080"
-    environment:
+    environ:ment
```
Swatinem, line 10: this looks like a breaking typo

Swatinem (approved):

docker-compose.yml
matt-codecov: how did i even do that

ty

PR #112, opened 2025-05-05
Improve cleanup progress reporting and scheduling

Swatinem:
This improves regular cleanup scheduling by integrating queryset-based cleanups with fn-based ones.

The progress reporting is changed in such a way that progress is logged regularly in a debounced fashion, as well as logging a summary once the contextmanager either finished regularly, or is being aborted due to a timeout.

### DIFF
--- a/apps/worker/services/cleanup/uploads.py
+++ b/apps/worker/services/cleanup/uploads.py
@@ -3,18 +3,16 @@ from datetime import datetime, timedelta, timezone
 
 from django.db.models.query import QuerySet
 
-from services.cleanup.cleanup import run_cleanup
-from services.cleanup.utils import CleanupContext, CleanupResult, CleanupSummary
+from services.cleanup.cleanup import cleanup_queryset
+from services.cleanup.utils import CleanupContext
 from shared.django_apps.reports.models import ReportSession as Upload
 
 UPLOAD_CHUNKSIZE = 5_000
 
 
-def cleanup_old_uploads(context: CleanupContext) -> CleanupSummary:
+def cleanup_old_uploads(context: CleanupContext):
     queries = create_upload_cleanup_queries()
     random.shuffle(queries)
-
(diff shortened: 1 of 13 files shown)


apps/worker/services/cleanup/utils.py
```diff
@@ -24,36 +44,42 @@ def __init__(self):
...
+        model_name = model.__name__ if model else self._current_model
+        if model_name and (cleaned_models or cleaned_files):
+            result = self.summary.summary.setdefault(model_name, CleanupResult(0))
+            result.cleaned_models += cleaned_models
+            result.cleaned_files += cleaned_files
+
+        now = time.monotonic()
+        if (now - self._last_progress_report) > 120.0:
+            self._last_progress_report = now
+            log.info("Cleanup is making progress…", extra={"summary": self.summary})
```
nora-codecov, line 71: nice

apps/worker/services/cleanup/uploads.py
```diff
@@ -24,11 +22,7 @@ def cleanup_old_uploads(context: CleanupContext) -> CleanupSummary:
                 break
 
             uploads_query = Upload.objects.filter(pk__in=upload_ids)
-            summary = run_cleanup(uploads_query, context=context)
-
-            complete_summary.add(summary)
-
-    return complete_summary
+            cleanup_queryset(uploads_query, context)
```
nora-codecov, line 25: The cleanups for `Owner` and `Repo` call this from within `cleanup_context()` and they add a `summary` to the `context` - did you mean to do it differently here? Why is it different from the other two?

nora-codecov (approved):

apps/worker/services/cleanup/uploads.py
Swatinem: primarily because the code and logs already existed, so no real reason :-)

PR #118, opened 2025-05-06
Remove all `labels` and `datapoints` related code

Swatinem:
Now that ATS has been deprecated, we can remove all the code related to `labels`, `label_index`, as well as `datapoints`.


Swatinem:
I think the perf regression in `test_report_carryforward` can be ignored. looking at the flamegraph diff, this looks like allocator / python runtime related, maybe a change in GC behavior.

On the other hand, IMO the ~10% improvement for the parsing (due to optimized `ReportLine` constructor) is more significant :tada:

giovanni-guidini (approved):

PR #119, opened 2025-05-06
Remove some unused code

Swatinem:
I discovered this while looking at our own code coverage.


apps/worker/services/lock_manager.py
```diff
@@ -47,12 +47,6 @@ def lock_name(self, lock_type: LockType):
         else:
             return f"{lock_type.value}_lock_{self.repoid}_{self.commitid}_{self.report_type.value}"
 
-    def is_locked(self, lock_type: LockType) -> bool:
```
matt-codecov, line 50: maybe we do want to use this? there is essentially duplicated locking code in `upload.py` (which is used in other tasks too iirc) https://github.com/codecov/umbrella/blob/5455e4b0850b82044c370fc54e8d6478b0efc9df/apps/worker/tasks/upload.py#L109-L111

Swatinem: yes, that code has its own lock implementation, and I inlined the `is_locked` method as it was only used in one place.

PR #120, opened 2025-05-06
build: update test results parser version

joseph-sentry:


apps/worker/uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 1
```
Swatinem, line 2: you can run `uv self update`, and then another `uv sync` to avoid having this churn in the lockfile

Swatinem (approved):

PR #122, opened 2025-05-06
fix: delete testruns with long strings

joseph-sentry:


Swatinem (approved):
sure, why not

Swatinem:
Some followup questions:
- Any idea why exactly timescale starts choking on large column values?
- Is this limit documented anywhere?
- `1_000` seems a bit arbitrary. I wonder if we could even tighten up this limit a bit more, maybe 256 or some such? Even that seems a bit excessive as test-id.

joseph-sentry:
@Swatinem 

> Any idea why exactly timescale starts choking on large column values?

Postgres errors with "Values larger than 1/3 of a buffer page cannot be indexed." when the timescale background job tries to update the continuous aggregate. It literally can't index those rows with a btree so it's either we try to change the type of the index or we set a limit on the size of the strings that we accept for rows that are indexed (or all strings in a testrun which is actually what we're doing) since if strings are too long that probably points to some sort of issue.

> Is this limit documented anywhere?

not really it's mentioned when you look up the error that we're getting: "btree version 4 maximum 2704 for index" and it just seems to be a fact of how btree indexes work since PG12

> 1_000 seems a bit arbitrary. I wonder if we could even tighten up this limit a bit more, maybe 256 or some such? Even that seems a bit excessive as test-id.

that's fair, it's completely arbitrary, given the limit was 2704 i thought 1000 was large enough for users that actually have long test names, but nowhere near the limit

PR #124, opened 2025-05-07
build: enable ta timeseries by default in devenv

joseph-sentry:
This option only really exists for self-hosted customers to manually toggle this feature in the case they don't want to deploy timescale, which will soon be required for TA. However when we're deploying locally, we always want to be running TA, so i'm enabling this by default in the devenv so that it works out of the box.


Swatinem (approved):

Swatinem:
The "umbrella secrets" already has this setting. The goal would be to have a passing testsuite locally without needing these secrets which  are not available for non-employees.

PR #127, opened 2025-05-08
fix: use correct invocation of migrate_timeseries.py in devenv startup

matt-codecov:
https://github.com/codecov/umbrella/pull/72 invokes this script in two different ways. this way is apparently wrong so i changed it to the other way


Swatinem (approved):

Swatinem:
I’m also wondering why it needs to be a separate script, instead of just integrating into the django manage infrastructure?

PR #129, opened 2025-05-08
Consolidate tests to use a unified `config/codecov.yml`

Swatinem:
This makes sure all the various docker images are using the same `config/codecov.yml`.

Alongside, this does fix up a few tests that were depending on some weird config behavior, as well as improve usage of mock configs.


apps/codecov-api/docker/test.yml
```diff
@@ -1,10 +1,16 @@
+# !!! ATTENTION !!!
```
matt-codecov, line 1: you could commit a symlink?

matt-codecov (approved):
when we run tests locally we use the already-running containers which have mounted the 1password `codecov.yml`. any thoughts on how we can get this `test.yml` into those containers and shim it into the test invocations?

apps/codecov-api/docker/test.yml
Swatinem: I haven’t actually tried, but all the docs are suggestion that docker *really* doesn’t like stuff being loaded/mounted from outside its `context`, also symlinks.

PR #139, opened 2025-05-12
add debugging to flare cleanup task

nora-codecov:
This task is silently failing every time it is initialized - adding some things to help us figure out what's going wrong.


Swatinem (approved):
lgtm.
the problem is not really that it is *failing*, but that it just doesn’t make any progress. I wonder if it might be stuck on one of those queries which might never complete.
I suggest running those with `explain (analyze)` to figure out if that is indeed the case.

PR #140, opened 2025-05-12
feat: show warnings from the test results parser in the PR comment

joseph-sentry:
I made a change to the test results parser to start returning parser warnings when something is wrong with the content of an uploaded JUnit XML, that the user should either report to us or look into fixing on their end, but should not affect the rest of the parsing. For example, the first type of warning is for test names being too long.
 

This PR is meant to take the warnings that the test results parser lib returns and show them on the PR comment, i also do some general cleaning up of tests and the notifier while i'm at it.

the high level is that we're creating the upload errors from the warnings in the processor task and we're adding the capability to display them in the notifier.

for the misc. changes i cleaned up the finisher tests, the notifier uses `<pre><code>` blocks instead of backticks now, and the finisher ignores the chain result


apps/worker/services/test_analytics/ta_processor.py
```diff
@@ -67,6 +68,11 @@ def ta_processor_impl(
             handle_parsing_error(upload, exc)
         return False
 
+    if update_state:
+        for info in parsing_infos:
+            for warning in info["warnings"]:
+                handle_warning(upload, warning)
```
Swatinem, line 74: can we rather bulk-insert these?

apps/worker/services/tests/snapshots/results__build_message__0.txt
```diff
@@ -0,0 +1,18 @@
+### :x: 1 Tests Failed:
+| Tests completed | Failed | Passed | Skipped |
+|---|---|---|---|
+| 3 | 1 | 2 | 3 |
+<details><summary>View the top 1 failed test(s) by shortest run time</summary>
+
+> <pre><code class="language-python">testname</code></pre>
```
Swatinem, line 7: does the markdown render properly when its nested in a `>`?

apps/worker/tasks/test_results_processor.py
```diff
@@ -112,6 +113,16 @@ def update_daily_totals(
         daily_totals[test_id]["skip_count"] += 1
 
 
+def handle_parsing_error(db_session: Session, upload: Upload, exc: Exception):
```
Swatinem, line 116: you have defined a fn for this one block, but are creating the `UploadError` objects inline further down below.
I believe you can move this inline as well.

apps/worker/tasks/test_results_processor.py
```diff
@@ -432,23 +445,26 @@ def process_individual_upload(
...
-        else:
+        for info in parsing_results:
+            for warning in info["warnings"]:
+                err = UploadError(
+                    report_upload=upload,
+                    error_code="warning",
+                    error_params={"warning_message": warning},
+                )
+                db_session.add(err)
+                db_session.commit()
```
Swatinem, line 465: I believe this should also use bulk insert.

Swatinem (approved):
bulk inserts would be nice.

I do like the snapshot tests, that looks much better than the inline tests before. Have you manually tested that the resulting markdown is formatted correctly?

apps/worker/services/tests/snapshots/results__build_message__0.txt
joseph-sentry: > <pre><code class="language-python"> yes</code></pre>

PR #142, opened 2025-05-13
Use more modern Django Middleware

michelletran-codecov:
Just a small change to stop using the deprecated middleware mixin.


apps/codecov-api/codecov_auth/middleware.py
```diff
@@ -7,9 +7,8 @@
     ACCESS_CONTROL_ALLOW_ORIGIN,
 )
 from corsheaders.middleware import CorsMiddleware as BaseCorsMiddleware
-from django.http import HttpRequest, HttpResponse
+from django.http import HttpRequest
 from django.urls import resolve
-from django.utils.deprecation import MiddlewareMixin
```
Swatinem, line 12: lol

apps/codecov-api/codecov_auth/middleware.py
```diff
@@ -45,10 +45,13 @@ class CurrentOwnerMiddleware(MiddlewareMixin):
     additional database queries).
     """
 
-    def process_request(self, request: HttpRequest) -> None:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
```
Swatinem, line 51: Out of curiosity, this looks a bit like the pattern is that you have one fn getting a `next` closure, expexting to return a closure with the same `Request -> Response` signature.

Is this class with init/call magic methods the intended way this is to be used, or should you actually rather have a fn that returns a closure?

Swatinem (approved):

PR #144, opened 2025-05-13
dedupe and clean up should_write_data_to_storage

nora-codecov:
`should_write_data_to_storage_config_check` was duplicated in api, worker and shared. The tests for it were duplicates, in api and worker.
removed the duplicates of `should_write_data_to_storage_config_check` in worker and api, keeping the one in shared.
Made a new test file for the code in shared, removed it from api and worker.

We also just had an issue with a value in `save_report_data_in_storage`, so I added a docstring and made the flow clearer. None of the logic changed, the tests (previously in api, now in libs/shared/tests/unit/utils/test_config.py) are unchanged.


apps/codecov-api/graphs/views.py
```diff
@@ -413,7 +413,7 @@ def get_pull_flare(self, pullid):
         pull = Pull.objects.filter(pullid=pullid, repository_id=repo.repoid).first()
         if pull is not None:
             # pull found
-            if pull._flare is not None or pull._flare_storage_path is not None:
+            if pull._flare or pull._flare_storage_path is not None:
```
michelletran-codecov, line 416: Can `_flare_storage_path` be an empty string? If not then we can simplify it:

```suggestion
            if pull._flare or pull._flare_storage_path:
```

michelletran-codecov (approved):

Swatinem (approved):
🎉

apps/worker/services/cleanup/pulls.py
```diff
@@ -28,105 +26,92 @@ def cleanup_flare(
...
+            log.info(
+                f"Flare cleanup: non_open_pulls_with_flare_in_db query {non_open_pulls_with_flare_in_db.explain(analyze=False)}"
+            )
+            batch_ids = list(
+                non_open_pulls_with_flare_in_db.values_list("id", flat=True)
+            )
+            # with analyze=True, the query is actually executed so cost and execution time are measured
+            log.info(
+                f"Flare cleanup: batch ids query {non_open_pulls_with_flare_in_db.values_list('id', flat=True).explain(analyze=True, verbose=True)}"
+            )
```
matt-codecov, line 57: strange that there doesn't seem to be a way to get the `EXPLAIN ANALYZE` result _and_ the results of the query in one go.

since you've just run the query a couple lines earlier, this run will be working with a warm cache and it will return faster. maybe run the `.explain(analyze=True)` one first

nora-codecov: oooooo i like that idea

apps/worker/services/cleanup/pulls.py
```diff
@@ -28,105 +26,92 @@ def cleanup_flare(
...
+            batch_of_id_path_pairs = list(
+                non_open_pulls_with_flare_in_archive.values_list(
+                    "id", "_flare_storage_path"
+                )
+            )
+            # with analyze=True, the query is actually executed so cost and execution time are measured
+            log.info(
+                f"Flare cleanup: batch_of_id_path_pairs query {non_open_pulls_with_flare_in_archive.values_list('id', '_flare_storage_path').explain(analyze=True, verbose=True)}"
+            )
```
matt-codecov, line 90: same note as before

matt-codecov (approved):
lgtm

PR #146, opened 2025-05-15
Defer merging line coverage to the serialization step

Swatinem:
When merging multiple uploads into one, it might be better to construct the per-line coverage from each session at the very end, instead of doing this on each merge.

This should help in particular when multiple uploads cover the same line, and we will go from calling `get_coverage_from_sessions` `O(N)` times for each upload/session to `O(1)`, making the whole operation `O(N)` instead of `O(N^2)` when we have `N` sessions to merge.

As a drive-by change, this removes the `messages` from the lines, as those are completely unused.


Swatinem:
Removal of `messages` is gaining us ~3% on some benchmarks which is below the report threshold.
The `O(N^2)` vs `O(N)` is not visible, as we are only merging the example report with itself (having the same sessionid), so the effect is not visible.

But I can also change the benchmark in such a way that the new code is being exercised.

apps/worker/test_utils/base.py
```diff
@@ -16,9 +15,7 @@ def convert_report_to_better_readable(self, report):
             file_report = report.get(filename)
             lines = []
             for line_number, line in file_report.lines:
-                (coverage, line_type, sessions, messages, complexity) = (
-                    dataclasses.astuple(line)
-                )
+                (coverage, line_type, sessions, messages, complexity) = line.astuple()
```
giovanni-guidini, line 18: the description says messages was removed?...

libs/shared/shared/reports/resources.py
```diff
@@ -270,7 +259,7 @@ def __contains__(self, filename):
         return filename in self._files
 
     @sentry_sdk.trace
-    def merge(self, new_report, joined=True):
+    def merge(self, new_report, joined=True, is_disjoint=False):
```
giovanni-guidini, line 262: I think a docstring explaining `is_disjoint` in this function would be nice, and also pointing out you need to call `finish_merge` at some point

(I know they are very close in the code)

libs/shared/shared/reports/totals.py
```diff
@@ -53,7 +49,7 @@ def get_line_totals(lines: Iterator[ReportLine]) -> ReportTotals:
         coverage=ratio(hits, total_lines) if total_lines else None,
         branches=branches,
         methods=methods,
-        messages=messages,
+        messages=0,
```
giovanni-guidini, line 52: messages is a number? unexpected...

apps/worker/test_utils/base.py
giovanni-guidini: hum looks like they were not _completely_ removed, but replaced with a constant, empty value. Makes sense to keep the interface I guess.

giovanni-guidini (approved):
LGTM, minor comments

apps/worker/test_utils/base.py
Swatinem: yes, because `complexity` is still used (although I believe we should remove that at some point as well), and we have to keep the order the same.

PR #149, opened 2025-05-15
Add JWT Middleware authentication with endpoint

michelletran-codecov:
This change adds a new middleware to process the JWT authentication for communicating with Sentry. It also adds a new GraphQL endpoint that uses this middleware for authentication (shouldn't affect existing endpoint).

This assumes the JWT claim to look something like:

```
{
  'g_u': 'github_service_id',
  'g_p': 'github',
}
```

### DIFF
--- a/apps/codecov-api/codecov_auth/middleware.py
+++ b/apps/codecov-api/codecov_auth/middleware.py
@@ -2,6 +2,7 @@ import logging
 from urllib.parse import urlparse
 
 import jwt
+from asgiref.sync import sync_to_async
 from corsheaders.conf import conf as corsconf
 from corsheaders.middleware import (
     ACCESS_CONTROL_ALLOW_CREDENTIALS,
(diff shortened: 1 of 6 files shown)


apps/codecov-api/codecov_auth/middleware.py
```diff
@@ -171,3 +174,63 @@ def middleware(request):
...
+                service_id=provider_user_id, service=provider
+            )
+            request.current_owner = owner
+            return await get_response(request, *args, **kwargs)
+
+        except jwt.ExpiredSignatureError:
+            log.warning(
+                "JWT token has expired",
+                extra={
+                    "token": token,
```
ajay-sentry, line 223: jw, what would logging the token allow us to do? And are there any security implications with logging the entire token?

ajay-sentry (approved):
lgtm! One comment on logging full token

apps/codecov-api/codecov_auth/middleware.py
michelletran-codecov: Hm... this is an expired token, so it might be OK? I can also just remove it just in case. Thanks for the catch!

PR #150, opened 2025-05-16
Implement consuming `vitest` json with `evals` metadata

Swatinem:
This is just a temporary solution to ingest and store evals metadata, before we have this properly standardized/implemented as a junit xml extension.
    
Eventually, we want to display this metadata as part of the PR comment.

---

The example snippet comes from https://github.com/getsentry/vitest-evals/issues/13#issuecomment-2835687888, first step of https://github.com/codecov/test-results-action/issues/120


giovanni-guidini (approved):
LGTM

PR #153, opened 2025-05-16
fix: use warning_message from warning upload errors

joseph-sentry:


ajay-sentry (approved):
lets clean this up on monday tho

PR #161, opened 2025-05-20
increase limits on flare_cleanup

nora-codecov:
Still gathering data on this task, so numbers are still hardcoded.
The last runs were items 1-5000 in batches of 50, and the whole thing finished in under 1 second, so I increased everything by a factor of 10.
Left one of the logs in as it was helpful to see timing and progress, especially if it fails midway through.


apps/worker/services/cleanup/tests/test_pulls.py
```diff
@@ -16,7 +16,7 @@ def test_flare_cleanup_in_regular_cleanup(
     mocker,
     mock_archive_storage,
 ):
-    mock_logs = mocker.patch("logging.Logger.info")
+    mocker.patch("logging.Logger.info")
```
Swatinem, line 19: you don’t need to mock the logger if you don’t want to assert calls again.

Swatinem (approved):
lgtm.
this conflicts partly with https://github.com/codecov/umbrella/pull/155

apps/worker/services/cleanup/tests/test_pulls.py
nora-codecov: when tests fail it will add the stdout to the output - the amount of logs generated for this range of ids is so high it struggles to load the action run. I thought about it and figured that i never want to see those logs in a unit test context (but I do still want them to log the actions when the real task runs in prod!). Mocking them this way prevents them from exploding whatever terminal is running the tests (if they happen to fail).

apps/worker/services/cleanup/pulls.py
```diff
@@ -112,12 +98,12 @@ def cleanup_flare(context: CleanupContext, batch_size: int = DELETE_FILES_BATCHS
                     _flare_storage_path=None
                 )
 
-            total_files_processed += len(batch_of_id_path_pairs)
-            context.add_progress(cleaned_files=len(successful_deletions), model=Pull)
+            cleaned_file_flares += len(successful_deletions)
+            context.add_progress(cleaned_files=cleaned_file_flares, model=Pull)
```
Swatinem, line 102:
```suggestion
            context.add_progress(cleaned_files=len(successful_deletions), model=Pull)
```

Otherwise it would over-count. Was this bug also present in my other PR?

Swatinem (approved):

PR #169, opened 2025-05-21
build: consolidate how tests are run between subprojects

matt-codecov:
build: consolidate how tests are run between subprojects


docker/Makefile.ci-tests
```diff
@@ -0,0 +1,66 @@
...
+		    --rootdir=${PYTEST_ROOTDIR}"
+
+# Outside the Docker env: Run integration tests.
+_test_env.run_integration:
+	docker-compose exec repo /bin/sh -c \
+		"COVERAGE_CORE=sysmon \
+		pytest --cov=${COV_SOURCE} \
+		    -m 'integration' \
+		    --cov-report=xml:unit.coverage.xml \
+		    --junitxml=unit.junit.xml \
```
Swatinem, line 55: the reason you see different coverage numbers might be related to this:
the integration test step is overwriting the unit test result files.

TBH, I still think we should stop splitting these two apart.
I believe the original reason to keep them was to dogfood flags. Which we do now as well because we have different sub-projects in the same repo, so uploads do exercise the flags/carryforward logic.

Swatinem (approved):

nora-codecov (approved):

PR #171, opened 2025-05-22
make app_id non-nullable

nora-codecov:
part 1: make this field non-nullable. 
Next step is to add a unique constraint


libs/shared/shared/django_apps/codecov_auth/migrations/migration_helpers/migration_0067.py
```diff
@@ -0,0 +1,34 @@
...
+    with transaction.atomic():
+        # group by insallation_id + app_id, getting the first object created from each unique group
+        # this will include objects with no duplicate AND the first created object in the set of duplicates
+        to_keep = (
+            model.objects.values("installation_id", "app_id")
+            .annotate(min_id=Min("id"))
+            .values_list("min_id", flat=True)
+        )
+
+        model.objects.exclude(id__in=to_keep).delete()
```
Swatinem, line 34: I would be curious what kind of query django generates for this.
Also django is notorious for going through python for these `delete`s. Which is the reason why I was rather using "raw delete" for the deletion job code.

Maybe it also makes sense here, depending on how many elements we end up hitting here.

libs/shared/shared/django_apps/codecov_auth/migrations/migration_helpers/migration_0067.py
```diff
@@ -0,0 +1,34 @@
+from django.db import transaction
+from django.db.models import Min
+
+from shared.config import get_config
+
+
+def backfill_app_id(model):
```
giovanni-guidini, line 7: Can't this code live directly in the migration file?

(obs.: idk if this is a pattern people use with django migrations?)

Swatinem (approved):
lgtm, though it would be nice to check ahead of time how expensive these queries will be.

giovanni-guidini (approved):
🐻: LGTM 

I think this is nice. We will most likely get some Sentry issues where the dupes are being created and then we can also fix that 😅

libs/shared/shared/django_apps/codecov_auth/migrations/migration_helpers/migration_0067.py
nora-codecov: I had it this way for testing - the test migrations class doesn't work that well, the only way to test the function that i could find quickly was to put the functions in their own file where I could call them directly 🥴

libs/shared/shared/django_apps/codecov_auth/migrations/migration_helpers/migration_0067.py
nora-codecov: I'm going to move this step into a separate migration and pr - but you're right, I'm going to try it this way first and if it's unable to complete I'll change it to a raw sql risky migration

libs/shared/shared/django_apps/codecov_auth/migrations/migration_helpers/migration_0067.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    Making this field non-nullable. It was nullable so that we could live load the app_id from yaml.
+    We no longer want to do that, so any GithubAppInstallations with None will be backfilled with our default app_id.
+    """
+    installation_default_app_id = get_config("github", "integration", "id")
+    if installation_default_app_id is not None:
```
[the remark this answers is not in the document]
nora-codecov: I threw this in there for dec and sh - I figured there's probably people running cc without any gh app business, so skip this for them since the table is probably empty. I added a note that if that field is not in configs, we assume there are no nulls that would need to be corrected in the db.

libs/shared/shared/django_apps/codecov_auth/migrations/migration_helpers/migration_0067.py
```diff
@@ -0,0 +1,34 @@
...
+        with transaction.atomic():
+            model.objects.filter(app_id__isnull=True).update(
```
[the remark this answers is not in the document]
nora-codecov: yes it does cause a seq scan - there are about 100k rows in the table, and 71 with null app_id

you're correct about the table size - it's small but accessed frequently, and this will cause a table lock. talked to trent about it and we thought this doesn't need to be handled as "risky" but should be done during off-peak hours, and I'll ask for an infra to be on hand in case it times out and needs to be retried

PR #179, opened 2025-05-27
build: unified pyproject.toml and requirements image

matt-codecov:
this PR:
- creates a unified `pyproject.toml`/`uv.lock` for everything in `umbrella`
- uses that `pyproject.toml`/`uv.lock` to build a single reqs image for all projects

all dependencies should be managed in the root `pyproject.toml` and `uv.lock` now. i think technically `shared` deps should be updated in `libs/shared/pyproject.toml` _in addition_ to the root one, but i think with how images are built we don't actually have to do that. i am going to try to make `shared` no longer have to be an actual package anyway.

dependencies are separated into groups:
- shared
- worker (includes shared)
- api (includes shared)
- dev
- prod (includes worker and api)
- everything (includes prod and dev)

the requirements image currently includes dev dependencies (which has always been true). changing that will be a separate matter.


docker/Makefile.docker
```diff
@@ -1,30 +1,40 @@
 export DOCKER_BUILDKIT=1
 
+export AR_REQS_REPO ?= ${AR_REPO_PREFIX}/umbrella-reqs
+export CI_REQS_REPO ?= codecov/umbrella-reqs
+
+# `LC_ALL=C` is added to `sort` to ensure you get the same order across systems.
+# Otherwise, a Mac may sort according to en_US.UTF-8 while CI may sort according to C/POSIX.
```
Swatinem, line 7: lol, I hate this behavior so so much. why are these coreutils just fundamentally "non-portable" in such a way :-(

pyproject.toml
```diff
@@ -3,10 +3,133 @@ name = "umbrella"
 version = "0.1.0"
 requires-python = ">=3.13"
 
+[tool.uv]
+default-groups = ["everything"] # TODO: Switch to "prod" when we can
```
Swatinem, line 7: nice!

apps/worker/pyproject.toml
```diff
@@ -1,75 +0,0 @@
...
-
-[build-system]
-requires = ["setuptools"]
-build-backend = "setuptools.build_meta"
-
-[tool.setuptools]
-py-modules = []
-
-[tool.uv]
-required-version = ">=0.7.5"
```
Swatinem, line 53: can you add this as well to the new central pyproject?
as `uv` had quite some churn recently what it is putting into the lockfile, and how those fields are named, we should make sure folks are on an uptodate version.

Swatinem (approved):
nice! I was battling really hard to make this a *workspace*, but seeing this now, it does make sense to actually have *one* project.

I’m afraid the tooling is still not really ready to have proper workspaces like they exist for Rust for example.

PR #180, opened 2025-05-27
build(compose): Minio port fixes and `README` improvements

ElioDiNino:


docker/codecov.yml
spalmurray, line 1: this feels wrong to me. What is this for?

PR #183, opened 2025-05-28
fix: remove auto-create indexes

joseph-sentry:
Timescale auto-creates indexes for each group by column paired with the time bucket column of a continuous aggregate, by default.

We don't want these indexes anymore because:
- they aren't used
- they're causing issues when users attempt to insert testruns with long names

the second condition made it so the job to recompute the CAgg from the base hypertable was failing:

because Postgres has a limitation on the size of entries in a BTree index, when we would try to add some row with a large name, it would fail to add an entry into the auto-created (name, timestamp_bin) index because the entry would exceed the limit imposed by Postgres.

The rows would be inserted into the hypertable with no problem because the hypertable didn't have an index on the name column.

This same problem applied to the other group by columns: testsuite, classname, and branch.

In the future, we may want to put an index on the base hypertable so the group by computation can use an index, and in that case we will want to change the group by condition to using the test_id, instead of the testsuite, classname and name.

Originally the test_id was meant to be a column that could be used to relate the testruns to objects outside of Timescale, like Flake objects in the test_analytics app, but it would fit the purpose of an index for the group by as well, since it's just a compressed representation of the (testsuite, classname, name) tuple.

The way the migration works is that it searches for all indexes related to the CAgg with the name in the where clause by joining the CAgg info table and the index info table. This currently deletes all indexes related to the CAgg, we will probably want to add some indexes back to improve query performance for when we query the CAgg in the cache rollups job.

### DIFF
--- a/libs/shared/shared/django_apps/ta_timeseries/migrations/0010_auto_20250528_1540.py
+++ b/libs/shared/shared/django_apps/ta_timeseries/migrations/0010_auto_20250528_1540.py
@@ -2,6 +2,8 @@
 
 from django.db import migrations
 
+from shared.django_apps.migration_utils import RiskyRunSQL
+
 
 class Migration(migrations.Migration):
     dependencies = [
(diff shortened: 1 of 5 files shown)


Swatinem (approved):
lgtm.
I assume this also drops any index on branch? do we still want that?

PR #192, opened 2025-05-29
add unique together constraint with dedupe logic

nora-codecov:
part 2: add unique constraint on app id and install id

app_id + installation_id should be unique on their own so I didn't get Owner involved because Owner is a bear.

part 1 is here https://github.com/codecov/umbrella/pull/171


ajay-sentry (approved):
send it

PR #193, opened 2025-05-29
fix: Redo SCoverage report parsing logic to parse external XML format

cchitneedi-ag:
The current parsers use the scoverage-data/scoverage.coverage.xml file which is meant to be used internally for storing data. SCoverage has a different XML reporting format for external consumption. Newer versions of the library default to txt files for internal data. SonarQube and other similar tools rely on the external format. The new implementation adds capability to parse the external format from XML attributes and computes partial lines based on the executed statements. Also adds new test case which is inline with the external format.

Changes are not breaking, the previous format is also supported.

Reference Links:
1. SCoverage XML Writer : https://github.com/scoverage/scalac-scoverage-plugin/blob/da4ca495dde236d6b52dd5e823a7d87324c74e57/reporter/src/main/scala/scoverage/reporter/ScoverageXmlWriter.scala#L73
2. SonarQube's expected file format : https://docs.sonarsource.com/sonarqube-server/latest/analyzing-source-code/test-coverage/test-coverage-parameters/#scala

### DIFF
--- a/apps/worker/services/report/languages/scoverage.py
+++ b/apps/worker/services/report/languages/scoverage.py
@@ -9,53 +9,68 @@ from .base import BaseLanguageProcessor
 from .helpers import child_text
 
 
+def get_child_value(element: Element, tag: str) -> str:
+    """
+    Returns the text of the child element with the given tag if it's a child of the given element.
+    Else if it's an attribute returns the attribute value instead
+    Else returns None
+    """
+    value = child_text(element, tag)
+    # If the child is not an empty string return  (It can't be None)
+    if value:
+        return value
+    return element.attrib.get(tag, "")
+
+
 class SCoverageProcessor(BaseLanguageProcessor):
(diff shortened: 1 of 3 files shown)


matt-codecov (commented):
thank you for the PR! it appears our CI is not working for community contributions (i might have broken it with a recent change) but we can get that sorted

left a few comments. i think i (or someone) will need to ramp up a bit on the format you're supporting here because there are some things i don't understand from a quick glance:
- what the `start` and `end` attributes on each `statement` mean
- why `line="5"` appears several times, some having `branch="true"` and others having `branch="false"`
- why something having partial coverage is not tied to the `branch` attribute
- why a fully-covered branch should be recorded as a line instead of a branch

does `/src/main/scala/CoverageClass.scala` have, like, 500-character lines with several totally distinct statements and control flow blocks on them?

if you could explain that'd help, else i'll try to get back to you early next week!

PR #194, opened 2025-05-29
feat(api): Authentication via Sentry GitHub App

ElioDiNino:
Passes down a flag from the middleware to appropriate functions touched by the GraphQL API when hitting the Sentry endpoint.

### DIFF
--- a/apps/codecov-api/services/tests/test_components.py
+++ b/apps/codecov-api/services/tests/test_components.py
@@ -108,7 +108,43 @@ class ComponentServiceTest(TestCase):
             ),
         ]
 
-        mock_final_yaml.assert_called_once_with(self.commit, user)
+        mock_final_yaml.assert_called_once_with(
+            self.commit, user, should_use_sentry_app=False
+        )
+
+    @patch("services.components.final_commit_yaml")
+    def test_commit_components_use_sentry_app(self, mock_final_yaml):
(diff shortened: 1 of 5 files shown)


apps/codecov-api/services/tests/test_components.py
```diff
@@ -108,7 +108,41 @@ def test_commit_components(self, mock_final_yaml):
...
+                        "flag_regexes": [r"flag.*"],
+                    },
+                    "individual_components": [
+                        {"component_id": "go_files", "paths": [r".*\.go"]},
+                    ],
+                }
+            }
+        )
+
+        user = AnonymousUser()
```
michelletran-codecov, line 131: We might want to put a comment that this only works because the `final_yaml` function is mocked out. The GItHub App _should_ be looked up via the user (so this can't technically be Anonymous if we tested this end to end).

michelletran-codecov (approved):
Generally LGTM!

Tagging @ajay-sentry and @matt-codecov as there's some changes to Sentry's Codecov client required (for the JWT claims, we removed `g_u` and adding `g_o` for the organization slug. I will make these changes tomorrow in Sentry unless someone gets to it today.

apps/codecov-api/services/tests/test_components.py
ElioDiNino: Great point, added!

PR #202, opened 2025-06-02
Add a new GraphQL mutation to sync with app token

michelletran-codecov:
This change makes the sync task use an app token rather than the user's token. It's a separate mutation, so it shouldn't conflict with the existing sync trigger.

Just also note that the `current_owner` is set to be the org with the JWT middleware: https://github.com/codecov/umbrella/blob/3541fdac533c9542505ab97d5ce5855db61278bd/apps/codecov-api/codecov_auth/middleware.py#L200.

### DIFF
--- a/apps/codecov-api/graphql_api/types/mutation/__init__.py
+++ b/apps/codecov-api/graphql_api/types/mutation/__init__.py
@@ -23,6 +23,7 @@ from .set_upload_token_required import gql_set_upload_token_required
 from .set_yaml_on_owner import gql_set_yaml_on_owner
 from .start_trial import gql_start_trial
 from .store_event_metrics import gql_store_event_metrics
+from .sync_repos import gql_sync_repos
 from .sync_with_git_provider import gql_sync_with_git_provider
 from .update_bundle_cache_config import gql_update_bundle_cache_config
 from .update_default_organization import gql_update_default_organization
@@ -33,6 +34,7 @@ from .update_self_hosted_settings import gql_update_self_hosted_settings
 mutation = ariadne_load_local_graphql(__file__, "mutation.graphql")
 mutation = mutation + gql_create_api_token
 mutation = mutation + gql_create_stripe_setup_intent
+mutation = mutation + gql_sync_repos
 mutation = mutation + gql_sync_with_git_provider
 mutation = mutation + gql_delete_session
 mutation = mutation + gql_set_yaml_on_owner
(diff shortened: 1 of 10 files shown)


apps/codecov-api/graphql_api/types/mutation/__init__.py
```diff
@@ -33,6 +34,7 @@
 mutation = ariadne_load_local_graphql(__file__, "mutation.graphql")
 mutation = mutation + gql_create_api_token
 mutation = mutation + gql_create_stripe_setup_intent
+mutation = mutation + gql_sync_repos
```
ajay-sentry, line 37: Thoughts on a comment here to let folks know these are both triggering the sync task, one is just using integration and the other isn't?

apps/codecov-api/graphql_api/types/mutation/sync_repos/sync_repos.py
```diff
@@ -0,0 +1,4 @@
+def resolve_sync_repos(_, info):
```
ajay-sentry, line 1: Do we also need to wrap this with the error handling decorator?
https://github.com/codecov/umbrella/blob/38f9e37b2a98b14958566f931da769ddb8f19be2/apps/codecov-api/graphql_api/helpers/mutation.py#L39-L47

Like in https://github.com/codecov/umbrella/blob/aa7bacd8b2a1615e76cae6a7e6b6d7be36812589/apps/codecov-api/graphql_api/types/mutation/sync_with_git_provider/sync_with_git_provider.py#L9

apps/codecov-api/graphql_api/types/mutation/sync_repos/sync_repos.graphql
```diff
@@ -0,0 +1,6 @@
+union SyncReposError = UnauthenticatedError
+
+type SyncReposPayload {
+  isSyncing: Boolean
```
ajay-sentry, line 4: nit: this should be required right?

ajay-sentry (approved):
lgtm, one small comment

apps/codecov-api/graphql_api/types/mutation/sync_repos/sync_repos.graphql
michelletran-codecov: Oh, it doesn't work with errors. I'll need to have that be nullable in case there is an error.

ajay-sentry: so we either return isSyncing _or_ the error in the payload

PR #208, opened 2025-06-03
chore: Remove old Codecov config files

ElioDiNino:
These are not used as far I as can tell (local tests and CI passes)


ElioDiNino:
Looks to be failing on the build step where it tries to remove the file 😆

PR #209, opened 2025-06-04
fix: Add build step to stage deploy workflow

ajay-sentry:
This PR adds the build step to the staging workflow, otherwise we don't have an image to reference (lol)


joseph-sentry (approved):
lgtm

PR #213, opened 2025-06-05
ci: change how push-stage.yml job is run

matt-codecov:
this PR makes `push-stage.yml` wait for `worker-ci.yml` and `api-ci.yml` to run before running. otherwise, it will trigger a whole bunch of duplicated image-building work:
- it will build a worker image while `worker-ci.yml` is also building one
- it will build an api image while `api-ci.yml` is also building one
- it will not wait on `_build-requirements.yml`, so both of the above images will also build (but not cache) the requirements image

this PR waits until `worker-ci.yml` and `api-ci.yml` have run before kicking off `push-stage.yml` so that it can reuse the images they built, and it uses `always()` to run regardless of whether the tests passed in those workflows

### DIFF
--- a/.github/workflows/ci-router.yml
+++ b/.github/workflows/ci-router.yml
@@ -4,6 +4,7 @@ on:
   push:
     tags:
       - prod-*
+      - stage-*
     branches:
       - main
       - staging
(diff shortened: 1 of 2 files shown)


ajay-sentry (approved):
lgtm!

PR #217, opened 2025-06-05
fix: handle new TA task failures gracefully

joseph-sentry:
Previously if the new TA processor was failing while in both mode, we would crash the entire test results processor task, but we want to have the old implementation still run even if the new one fails, so this fixes that.


Swatinem (approved):

JerrySentry:
bugbot run

PR #218, opened 2025-06-05
[fix] Add github service filter check to Gen AI auth

rohitvinnakota-codecov:
We use this endpoint to validate that a user has the Codecov AI app installed before initiating any Seer workflows. It's possible for multiple owners to have the same service ID so there were cases where we were not fetching the right owner. This change fixes that.


JerrySentry:
bugbot run

spalmurray (approved):

PR #219, opened 2025-06-09
Dedupe GithubAppInstallation and add unique constraint

nora-codecov:
Previous attempt at simple migration failed - need to do risky migration with raw sql commands.
previous (reverted) attempt [here](https://github.com/codecov/umbrella/pull/192) 

Also, had to sort out foreign key connections to CommitNotifications. Since the duplicate GithubAppInstallation objects mean the same thing, my script transfers any attached to a duplicate to a keeper, then deletes the duplicate.

Tried writing unit tests but could not get `TestMigrations` to work to save my life, so did manual testing locally, everything works as intended 🚀 🚀 🚀


JerrySentry:
bugbot run verbose=true

jason-ford-codecov (approved):

PR #223, opened 2025-06-10
refactor: move get_settings_module django helper to shared

matt-codecov:
refactor: move get_settings_module django helper to shared


Swatinem (approved):

libs/shared/shared/django_apps/utils/config.py
```diff
@@ -1,29 +1,33 @@
...
-    settings_module = SettingsModule.DEV.value
-elif RUN_ENV == "STAGING":
-    settings_module = SettingsModule.STAGING.value
-elif RUN_ENV == "TESTING":
-    settings_module = SettingsModule.TESTING.value
-elif RUN_ENV == "ENTERPRISE":
-    settings_module = SettingsModule.ENTERPRISE.value
-else:
-    settings_module = SettingsModule.PRODUCTION.value
+def get_settings_module(parent_module: str = "codecov") -> str:
```
ElioDiNino, line 13: Regarding the build failure which applies to `_settings_module` in this function: https://stackoverflow.com/questions/74412503/cannot-access-local-variable-a-where-it-is-not-associated-with-a-value-but

ElioDiNino (changes requested):

libs/shared/shared/django_apps/utils/config.py
matt-codecov: lol lint got mad at me for declaring `global _settings_module` but i guess i have to do that after all

ElioDiNino (approved):

PR #237, opened 2025-06-16
fix: CORS custom logic returns `*` for external origins

ElioDiNino:
Our API currently adds the `Access-Control-Allow-Credentials` header to requests originating from our whitelisted domains and sets `Access-Control-Allow-Origin` to the request origin for **all requests**. This is misleading since it implies every origin is on our whitelist and it risks us accidentally setting `Access-Control-Allow-Credentials` on non-whitelisted origins (whereas setting `Access-Control-Allow-Credentials` is a specification violation if `Access-Control-Allow-Origin` is set to `*`). To this point, we have received two security reports about our CORS configuration, where the researchers assume we have unintentionally allowed requests from every domain.

This PR fixes the issue by leaving credentials and standard origin headers to the base CORS module and then based on a global setting, will set `Access-Control-Allow-Origin` to `*` for all external origins if enabled.

Closes https://github.com/codecov/engineering-team/issues/2850

### DIFF
--- a/apps/codecov-api/codecov/settings_dev.py
+++ b/apps/codecov-api/codecov/settings_dev.py
@@ -27,8 +27,7 @@ CODECOV_DASHBOARD_URL = get_config(
 
 CORS_ALLOWED_ORIGINS = [
     CODECOV_DASHBOARD_URL,
-    "http://localhost",
-    "http://localhost:9002",
+    "http://localhost:9002",  # Minio port when not using gateway
 ]
 
 COOKIES_DOMAIN = "localhost"
--- a/apps/codecov-api/codecov/settings_dev.py
+++ b/apps/codecov-api/codecov/settings_dev.py
@@ -19,7 +19,6 @@ STRIPE_ENDPOINT_SECRET = get_config(
     "services", "stripe", "endpoint_secret", default="default"
 )
 
-CORS_ALLOW_CREDENTIALS = True
 
 CODECOV_URL = "localhost"
 CODECOV_API_URL = get_config("setup", "codecov_api_url", default=CODECOV_URL)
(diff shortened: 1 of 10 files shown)


tools/devenv/config/test.yml
```diff
@@ -1,5 +1,5 @@
 setup:
-  codecov_url: https://codecov.io
+  codecov_url: http://localhost:3000
```
spalmurray, line 2: why this change? I'd imagine it doesn't really matter but just curious

apps/codecov-api/codecov/settings_dev.py
```diff
@@ -19,16 +19,16 @@
     "services", "stripe", "endpoint_secret", default="default"
 )
 
-CORS_ALLOW_CREDENTIALS = True
-
-CODECOV_URL = "localhost"
+CODECOV_URL = get_config("setup", "codecov_url", default="http://localhost:8080")
 CODECOV_API_URL = get_config("setup", "codecov_api_url", default=CODECOV_URL)
-CODECOV_DASHBOARD_URL = "http://localhost:3000"
+CODECOV_DASHBOARD_URL = get_config(
```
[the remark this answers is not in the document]
ElioDiNino: Haha yup, I've been having the same issue and it was bugging me

tools/devenv/config/test.yml
spalmurray: Oh I think it's the thing in `apps/codecov-api/codecov_auth/tests/unit/test_middleware.py`? Leave it if you want, but nit is maybe not using 3000 as we don't really use it anymore for gazebo.

spalmurray (approved):
Couple comments. All nits though. 👍 from me

tools/devenv/config/test.yml
ElioDiNino: Now that the dev settings use `codecov_url` to override `codecov_dashboard_url` instead of ignoring it, running the tests locally was failing since some of the API tests expect the URL to be `http://localhost:3000`

ElioDiNino: Pushed a change to modify the tests instead

PR #239, opened 2025-06-16
fix: Override default Minio port and use rolling gateway image

ElioDiNino:
The default Minio port is 9000, so we need to override it. Additionally, we want our development environment to use the latest gateway version, instead of waiting for explicit releases.


matt-codecov (approved):
i guess using `rolling` is probably better, good idea

PR #246, opened 2025-06-18
POC: endpoint to get evals info

giovanni-guidini:
These changes add 2 new endpoints in the REST api (hidden behing a feature flag) with the objective of being a POC in reporting Eval info saved in Test Analytics.

Probably not what anyone else had in mind, but a POC mind you. I didn't want to deal with building a frontend. Also the current way _we_ use evals is not in the CI, so I didn't think it made much sense to mess around the PR comment either.

Anyway the endpoints are:
* summary - aggregates eval data (that is saved per test-run).
* compare - aggregate per commit and compare.

> **:warning:** I had to change the DB router for the API for tests to pass.
> It included a custom field lookup that was moved to the shared DB router.

### DIFF
--- a/apps/codecov-api/api/public/v2/evals/views.py
+++ b/apps/codecov-api/api/public/v2/evals/views.py
@@ -17,12 +17,12 @@ from shared.django_apps.ta_timeseries.models import Testrun
 
 
 class EvalsSummary(TypedDict):
-    avgDuration: float
+    avgDurationSeconds: float
     avgCost: float
     totalItems: int
     passedItems: int
     failedItems: int
-    scores: dict[str, float]
+    scores: dict[str, dict[str, float]]
 
 
 class EvalsFilters(django_filters.FilterSet):
@@ -90,8 +90,8 @@ class EvalsViewSet(viewsets.GenericViewSet, RepoPropertyMixin):
             else 0
         )
 
-        # Calculate score averages for passed items
-        score_avgs_counts: dict[str, tuple[float, int]] = {}
+        # Calculate score sums and averages for all items with scores
+        score_agg_data: dict[str, tuple[float, int]] = {}
         cost_acc = 0
         items_with_cost = 0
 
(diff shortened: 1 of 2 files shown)


Swatinem (approved):
lgtm, but I would also recommend a review from someone else who knows this db router things better than me.

apps/codecov-api/api/public/v2/evals/views.py
```diff
@@ -0,0 +1,233 @@
...
+    @extend_schema(
+        summary="Evaluation summary",
+        parameters=[
+            OpenApiParameter(
+                "commit",
+                OpenApiTypes.STR,
+                OpenApiParameter.QUERY,
+                description="commit SHA for which to return evaluation summary",
+            ),
+            OpenApiParameter(
```
JerrySentry, line 146: what does the `classname` param mean?

apps/codecov-api/api/public/v2/evals/views.py
```diff
@@ -0,0 +1,233 @@
...
+
+class EvalsSummary(TypedDict):
+    avgDuration: float
+    avgCost: float
+    totalItems: int
+    passedItems: int
+    failedItems: int
+    scores: dict[str, float]
+
+
```
JerrySentry, line 27: are the eval reports sharing the same table `Testrun` as TA related reports? How does the queryset only pull eval reports?

apps/codecov-api/api/public/v2/evals/views.py
```diff
@@ -0,0 +1,233 @@
...
+        passed_items = sum(1 for t in testruns if t.outcome == "pass")
+        failed_items = total_items - passed_items
+
+        avg_duration = (
+            sum(t.duration_seconds or 0 for t in testruns) / total_items
+            if total_items > 0
+            else 0
+        )
+
+        # Calculate score averages for passed items
```
JerrySentry, line 93: I think it would be useful to have the failed items have a score of 0 instead of being ignored. Maybe a separate metric for that. Because just looking at a 100% score without looking at number of failed items seems deceiving, unless it failed because the evals code errored. Ideally we have 1 number to paint as much picture as possible.

apps/codecov-api/api/public/v2/evals/views.py
```diff
@@ -0,0 +1,233 @@
...
+from api.public.v2.schema import repo_parameters
+from api.shared.mixins import RepoPropertyMixin
+from api.shared.permissions import RepositoryArtifactPermissions
+from rollouts import READ_NEW_EVALS
+from shared.django_apps.db_settings import TA_TIMESERIES_ENABLED
+from shared.django_apps.ta_timeseries.models import Testrun
+
+
+class EvalsSummary(TypedDict):
+    avgDuration: float
```
suejung-sentry, line 20: nit: can consider including the units in the name (or at least in a comment). Is it avgDurationSeconds?

apps/codecov-api/codecov/db/__init__.py
```diff
@@ -1,69 +0,0 @@
-import logging
-
-from django.conf import settings
-from django.db.models import Field, Lookup
-
-log = logging.getLogger(__name__)
-
-
-class DatabaseRouter:
```
suejung-sentry, line 9: I quickly skimmed this against the MultiDatabaseRouter - there are small differences in references to the `ta_timeseries` db - is that okay to cut over to? (I'm not very familiar with the topology here)

suejung-sentry (approved):

apps/codecov-api/api/public/v2/evals/views.py
giovanni-guidini: This comes from the [Testrun model](https://github.com/codecov/umbrella/blob/30767a3a07b40725fb5e1e2c411640ba04809b69/libs/shared/shared/django_apps/timeseries/models.py#L196). I believe in the test definitions where you have `class TestRelatedTests` it would the that class name. In vitest it is the `describe` name.

What matters to this POC is that [ta_processing](https://github.com/codecov/umbrella/blob/30767a3a07b40725fb5e1e2c411640ba04809b69/apps/worker/services/test_analytics/ta_processor.py#L139) maps that to the "ancestorTitle" (meaning the class name I described above). For langfuse data it is the run name, basically.

I do agree it's a bad name and if we are to use this for real we'd need to change it to something better.

I'll add a comment explaining what it is in the code.

PR #248, opened 2025-06-18
feat: add total duration column to test results

joseph-sentry:
we want to add the total duration for each test to the test results so we can show users granular information about their test performance. This is because it's valuable for users to have granular information about the total impact of each test, since right now we only show a high level view of the total duration of their tests.

So I am:
- Adding a new field to the GQL interface
- Populating it by computing the total duration in the polars dataframe by doing: `avg duration * (pass count + fail count)
- adding a test to check if we can order by the total duration properly

As an aside I'm:
- renaming some confusing function and test names
- factoring out some repeated expressions in the polars df aggregation


apps/codecov-api/graphql_api/tests/test_test_analytics.py
```diff
@@ -772,7 +778,8 @@ def test_gql_query_test_suites_term(self, repository, store_in_redis):
             "testsuite1",
         ]
 
-    def test_gql_query_with_new_ta_v1(self, mocker, repository, snapshot):
+    @pytest.mark.parametrize("ordering", ["FAILURE_RATE", "TOTAL_DURATION"])
+    def test_gql_query_with_new_ta_v1(self, mocker, repository, snapshot, ordering):
```
ajay-sentry, line 782: this name is a code smell if we can find a better name for that

apps/codecov-api/utils/test_results.py
```diff
@@ -198,6 +209,10 @@ def v1_agg_table(table: pl.LazyFrame) -> pl.LazyFrame:
         pl.col("avg_duration") * (pl.col("pass_count") + pl.col("fail_count"))
     ).sum() / (pl.col("pass_count") + pl.col("fail_count")).sum()
 
+    total_duration_expr = (
```
ajay-sentry, line 212: And here

ruff.toml
```diff
@@ -29,4 +29,8 @@ ignore = [
     # Trailing whitespace, and blank lines which we use in test assertions:
     "W291",
     "W293",
+    "PLC0415",
```
ajay-sentry, line 32: How come we're also updating Ruff in this PR? Anyway we can split that stuff out or revert these changes?

apps/codecov-api/utils/test_results.py
```diff
@@ -186,17 +184,16 @@ def no_version_agg_table(table: pl.LazyFrame) -> pl.LazyFrame:
 
 
 def v1_agg_table(table: pl.LazyFrame) -> pl.LazyFrame:
```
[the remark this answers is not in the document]
joseph-sentry: `v1` is my attempt at doing some versioning of the rollup dataframe schema.

I will remove the rollup file soon, so this will be going away, but if we were to keep it, this would be handy if we ever wanted to change the dataframe schema.

PR #249, opened 2025-06-18
chore: Resolve new ruff checks

ElioDiNino:
* Added explicit ignores where circular imports were a problem
  * Move imports to the top of the file where it wasn't
* Ran ruff `--fix` for the generic type fixes
* Implemented missing hash functions for classes that implement `__eq__`


giovanni-guidini (approved):
nice 🥔

PR #259, opened 2025-06-23
feat: Celery task to run `pgpartition` on a schedule

ElioDiNino:
This Celery task ensures that partitions are automatically managed (created and deleted) instead of relying on startup scripts running on new deploys.

### DIFF
--- a/apps/worker/tasks/partition_management.py
+++ b/apps/worker/tasks/partition_management.py
@@ -1,0 +1,55 @@
+import logging
+import sys
+from io import StringIO
+
+from django.core.management import call_command
+
+from app import celery_app
+from shared.celery_config import partition_management_task_name
+from tasks.crontasks import CodecovCronTask
+
+log = logging.getLogger(__name__)
+
+
+class PartitionManagementTask(CodecovCronTask, name=partition_management_task_name):
+    """
+    Task to manage database partitions by running the pgpartition command.
+    This handles creating future partitions and deleting old partitions
+    based on the configuration in partitioning.py.
+    """
+
+    @classmethod
+    def get_min_seconds_interval_between_executions(cls):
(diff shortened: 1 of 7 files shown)


apps/worker/tasks/partition_management.py
```diff
@@ -0,0 +1,44 @@
...
+class PartitionManagementTask(CodecovCronTask, name=partition_management_task_name):
+    """
+    Task to manage database partitions by running the pgpartition command.
+    This handles creating future partitions and deleting old partitions
+    based on the configuration in partitioning.py.
+    """
+
+    @classmethod
+    def get_min_seconds_interval_between_executions(cls):
+        return 18000  # 5h
```
joseph-sentry, line 22: is this 5 hours somehow related to the 0500 in the crontab or is that just a coincidence

ElioDiNino: Purely coincidence and I chose the value fairly arbitrarily other than preventing the task from being able to run too many times in a day. Open to other value suggestions

apps/worker/tasks/partition_management.py
```diff
@@ -0,0 +1,55 @@
...
```
michelletran-codecov, line 23: Can we extract this into a constant? I see that the 5 hours is also referenced in the celery_config. I would suggest extracting this into a settings variable and reusing that in both locations to ensure consistency.

ElioDiNino: The values are unrelated, see: https://github.com/codecov/umbrella/pull/259#discussion_r2162472360

michelletran-codecov: Oh, whoops! Should've checked collapsed messages!

PR #264, opened 2025-06-25
fix: add cutoff date to TA dashboard read path

joseph-sentry:
I added functionality in the worker previously to enable the "new" impl_type for repos that started uploading after a specific cutoff date since we didn't need to wait for data to backfill into timescale.

I failed to match that change here, so users were only writing to timescale on the worker side but trying to read from the old rollups on the API side.

this commit fixes this issue by checking the cutoff date alongside the feature flag.

### DIFF
--- a/apps/codecov-api/utils/test_results.py
+++ b/apps/codecov-api/utils/test_results.py
@@ -152,12 +152,7 @@ def old_get_results(
             # give up
             return None
 
-    # deserialize - result should be bytes at this point
-    if isinstance(result, bytes):
-        table = pl.read_ipc(result)
-    else:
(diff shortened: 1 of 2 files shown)


apps/codecov-api/utils/test_results.py
```diff
@@ -115,7 +136,7 @@ def old_get_results(
 ) -> pl.DataFrame | None:
     redis_conn = get_redis_connection()
     key = redis_key(repoid, branch, interval_start, interval_end)
-    result: bytes | None = redis_conn.get(key)
```
ajay-sentry, line 118: jw whyd we get rid of the type here

PR #267, opened 2025-06-26
Add a Sentry ID field and update Account Plans

michelletran-codecov:
These changes are required to support Codecov Merge. In particular:

* Adding a Sentry Org ID to associate the account to a specific Sentry Org
* Updating the plans to include a Sentry Merge type plan

The accounts table is small, so we don't need a risky migration.

### Legal Boilerplate


libs/shared/shared/plan/constants.py
```diff
@@ -43,6 +43,7 @@ class PlanName(enum.Enum):
     ENTERPRISE_CLOUD_MONTHLY = "users-enterprisem"
     ENTERPRISE_CLOUD_YEARLY = "users-enterprisey"
     USERS_DEVELOPER = "users-developer"
+    SENTRY_MERGE_PLAN = "sentry-merge"
```
ajay-sentry, line 46: Could we add a comment here that this is only used for the account model?

ajay-sentry (approved):

PR #269, opened 2025-06-27
f

spalmurray:
### Legal Boilerplate


libs/shared/shared/django_apps/codecov_auth/models.py
```diff
@@ -121,6 +121,20 @@ def is_github_student(self) -> bool:
             return False
         return github_owner.student
 
+    @property
+    def default_org(self):
+        org: Owner = self.owners.first()
```
ajay-sentry, line 126: What do you think about 

try :
return org.default_org.username
catch AttributeError:
return None

I forget the exact exception but reads a little neater imo

spalmurray: yea that's fair

PR #275, opened 2025-07-03
fix: take flaky_fail_count into account for total tests in new TA impl

joseph-sentry:
in the old TA implementation the possible outcomes were: pass, fail, and skip.
flaky failures were a subset of fails which meant we shouldn't include the
flaky_fail_count in the total tests calculation

In the new implementation that's not the case, flakiness is represented by the
outcome of a testrun, which means that we have to count flaky fails in the total
tests

This commit also does some light refactoring of the polars aggregation
expressions for simplicity, and fixes a bug where we weren't running fill_nan
when it was needed.

### DIFF
--- a/apps/codecov-api/graphql_api/types/test_results_aggregates/test_results_aggregates.py
+++ b/apps/codecov-api/graphql_api/types/test_results_aggregates/test_results_aggregates.py
@@ -4,56 +4,59 @@ from graphql import GraphQLResolveInfo
 
 from graphql_api.types.enums.enum_types import MeasurementInterval
 from utils.ta_types import TestResultsAggregates
-from utils.test_results import get_results
+from utils.test_results import get_results, use_new_impl
 
 
-def calculate_aggregates(table: pl.DataFrame) -> pl.DataFrame:
-    return table.with_columns(pl.col("avg_duration").fill_nan(0)).select(
-        (
-            pl.col("avg_duration")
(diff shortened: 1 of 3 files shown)


ElioDiNino (approved):

apps/codecov-api/graphql_api/types/test_results_aggregates/test_results_aggregates.py
```diff
@@ -23,48 +23,53 @@ class TestResultsAggregates:
...
+        total_duration,
+        slowest_tests_duration,
         (pl.col("total_skip_count").sum()).alias("skips"),
         (pl.col("total_fail_count").sum()).alias("fails"),
-        (
-            (pl.col("avg_duration") >= pl.col("avg_duration").quantile(0.95))
-            .top_k(min(100, max(table.height // 20, 1)))
-            .sum()
-        ).alias("total_slow_tests"),
+        pl.lit(num_slow_tests).alias("total_slow_tests"),
```
ajay-sentry, line 54: @cursor what does .lit do here

apps/codecov-api/utils/test_results.py
```diff
@@ -93,6 +98,17 @@ def _has_commits_before_cutoff(repoid: int) -> bool:
     ).exists()
 
 
+@lru_cache(maxsize=1000)
```
ajay-sentry, line 101: why 1000? Is the default of 128 not good enough for us?

PR #278, opened 2025-07-03
fix(shared): Legacy migrations warning

ElioDiNino:
Django currently thinks the index is named `yaml_histor_ownerid_74e79b_idx` but due to migration 0002 it’s actually saved in the database as `yaml_history_ownerid_timestamp`. This change brings them in sync so Django no longer wants to incorrectly make a new migration to change it to a different sha-based index name.


michelletran-codecov (approved):
LGTM, but thoughts on removing this table altogether? It doesn't look like the `YamlHistory` is used at all (staging has a few entries, but nothing else in other DB). Also code-wise, looks like a safe removal of the model as it's not really referenced in any features.

ElioDiNino:
> LGTM, but thoughts on removing this table altogether? It doesn't look like the `YamlHistory` is used at all (staging has a few entries, but nothing else in other DB). Also code-wise, looks like a safe removal of the model as it's not really referenced in any features.

Works for me. The `legacy_migrations` Django app is not installed in prod which is why the table isn't there so should be safe to delete.

PR #285, opened 2025-07-04
feat: create and add sane_repr to TA models

joseph-sentry:
Sentry has this in their repo and I think it's pretty nifty so I'm adding it here for now


ElioDiNino (approved):
Unit tests for this might be good but implementation looks 👍

PR #298, opened 2025-07-07
fix(billing): Fix update address raise error

suejung-sentry:
When I tried to update the billing address of Codecov org in prod (update the street name from "St." to "St"), in the UI it made it look like it went through successfully and the PATCH /update-billing-address 200'd.

But then going through the [logs](https://cloudlogging.app.goo.gl/Up12cj6WUXWnGbbQ6) showed the error 
`"Request req_9ak7WzhcpRmZgQ: Your card was declined."`
And the address is not reflected as changed within Stripe (and therefore not on our UI either).

Fix in this PR so this error can be passed back to gazebo frontend.

Screenshot 2025-07-07 at 2 24 38 PM


ajay-sentry:
@suejung-sentry How does this look on the UI when an error is raised?

apps/codecov-api/services/billing.py
```diff
@@ -793,6 +792,7 @@ def update_billing_address(self, owner: Owner, name, billing_address):
                     "error_type": type(e).__name__,
                 },
             )
+            raise e
```
ajay-sentry, line 795: Nice find! Also wondering if we should add this to the update_email_address function as well :o

suejung-sentry: I looked over that one and I couldn't figure out a way to trigger that error scenario with stripe. I'm also hesitant to touch it since will probably need to go fix it up gazebo side too & we haven't heard of this being much of an issue. As in at this point I feel like it's more likely I'll make more work for someone by changing it than leaving it be

apps/codecov-api/services/tests/test_billing.py
```diff
@@ -1859,20 +1859,29 @@ def test_update_email_address(self, modify_customer_mock):
...
+    @patch("services.billing.stripe.PaymentMethod.modify")
+    @patch("services.billing.stripe.Customer.modify")
+    def test_update_billing_address_with_invalid_address(
+        self,
+        modify_customer_mock,
+        modify_payment_mock,
+        retrieve_customer_mock,
+    ):
         owner = OwnerFactory(stripe_customer_id="123", stripe_subscription_id="123")
-        assert self.stripe.update_billing_address(owner, "John Doe", "gabagool") is None
```
suejung-sentry, line 1865: this test was failing in CI dunno, was giving error about missing API key again. So changed it to take mocks for the calls to stripe api and still tests invalid address scenario

.vscode/extensions.json
```diff
@@ -2,6 +2,7 @@
   "recommendations": [
     "ms-python.python",
     "ms-python.mypy-type-checker",
+    "elagil.pre-commit-helper",
```
suejung-sentry: decided to remove this as this fixed it for me but I'm second guessing whether it was just a me-problem and don't want to impose it on the world

ajay-sentry: hahaha fair enough

PR #301, opened 2025-07-09
fix: Add caps to celery work

suejung-sentry:
[UPDATE - this PR as is won't work unless I keep track of the start somewhere durable.. thinking.. tbd]

[This PR](https://github.com/codecov/umbrella/pull/232) removed bounds from cleanup_flare() function, causing it to process potentially millions of records instead of limited batches. This created too many concurrent file/socket connections (redis, postgres, GCS), exceeding the system's file descriptor limit (ulimit), resulting in "Error 24: Too many open files" in Celery.

Fix: Added back bounded processing (max 200 batches per run), used .iterator() for memory efficiency, and added throttling (time.sleep(0.005)) between file deletions to prevent connection spikes. Note that this means it will take more daily jobs to pay down the rows needing cleanup, but we'll do it in more manageable batches.

Also note this PR retains the changes that the original above PR was meant to solve (i.e., look at everything instead of just ids # 5000-500000, so we correctly process 1-4999 and 500001+).

Closes https://linear.app/getsentry/issue/CCMRG-1352/fix-ongoing-intermittent-celery-task-failures-reported-issue

### DIFF
--- a/apps/worker/services/cleanup/pulls.py
+++ b/apps/worker/services/cleanup/pulls.py
@@ -1,8 +1,8 @@
 import logging
+import time
 
 import sentry_sdk
 from django.db.models.aggregates import Max
-
 from services.cleanup.models import DELETE_FILES_BATCHSIZE
 from services.cleanup.utils import CleanupContext
 from shared.django_apps.core.models import Pull, PullStates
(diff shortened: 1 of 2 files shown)


apps/worker/services/cleanup/pulls.py
```diff
@@ -1,107 +1,115 @@
...
+    file_batch_size: int = DELETE_FILES_BATCHSIZE,
+    max_total_batches: int = 200,  # Cap work per run
 ):
     """
     Flare is a field on a Pull object.
-    Flare is used to draw static graphs (see GraphHandler view in api) and can be large.
-    The majority of flare graphs are used in pr comments, so we keep the (maybe large) flare "available"
-    in Archive storage while the pull is OPEN.
+    
+    If the pull is OPEN, we keep the flare data for quick access.
```
suejung-sentry, line 25: TODO: fix my IDE, the linting is too aggressive rn

PR #312, opened 2025-07-14
ref: GQL query_to_connection to allow multiple ordering directions

joseph-sentry:
I want to be able to specify different ordering directions for each ordering parameter


ElioDiNino (approved):

ajay-sentry:
> an example is in the TA GQL pagination i want to order by whatever column the user specifies, in the direction they specify but always order by name ascending second.

Is this a customer ask or something that we wanted to do just cuz? Like @spalmurray mentioned, we should definitely test that setting ordering to be a () from None doesn't break existing pagination elsewhere.

And if its the latter, and this ends up blowing up into a larger thing maybe we can revisit if the juice is worth the squeeze at all

joseph-sentry:
> Is this a customer ask or something that we wanted to do just cuz

more of a "just cuz" but also because i think it'll make the resulting product more polished.

>  should definitely test that setting ordering to be a () from None doesn't break existing pagination elsewhere

i don't think it will because the previous code implicitly relied on ordering being an iterable since it was doing a `field in ordering` tuple comprehension type thing without changing ordering. the ordering change in this PR is just a typing fix

the bulk of this change (and imo the most "risky" part) is in the ordering_direction now being allowed to be an iterable of `OrderingDirection` alongside its previous types, but we know that the previous code assumed the `ordering_direction` to either be `None` or `OrderingDirection`, and i'm explicitly handling the case where it isn't either of those things. 

I just made another change to make sure we're handling the `None` case the same way we used to even though by my observation we're not calling `query_to_connection` with `ordering_direction=None` anywhere in the codebase

apps/codecov-api/graphql_api/helpers/connection.py
```diff
@@ -260,7 +261,19 @@ def queryset_to_connection_sync(
         return ArrayConnection(array_paginator)
 
     else:
-        ordering = tuple(field_order(field, ordering_direction) for field in ordering)
+        if ordering_direction is None:
```
ajay-sentry, line 264: nice

PR #314, opened 2025-07-14
chore: Migrations refactor

ElioDiNino:
* Adds the migrate scripts to both worker and API, and keep them in sync
* API runs timeseries migrations via the safer script (versus the unsafe migrate command)
* Rename devenv command for better correctness

### DIFF
--- a/apps/worker/migrate_timeseries.py
+++ b/apps/worker/migrate_timeseries.py
@@ -7,8 +7,17 @@ from django.db import connections
 
 from shared.django_apps.utils.config import get_settings_module
 
+#################################################################
+# Keep in sync between apps/codecov-api/migrate_timeseries.py and
+# apps/worker/migrate_timeseries.py
+
+# Note: this is the only difference from the worker version
+parent_module = "django_scaffold"
+#################################################################
+
+
 # Setup Django environment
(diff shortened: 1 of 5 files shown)


apps/worker/migrate_timeseries.py
```diff
@@ -17,29 +26,35 @@
...
-            with connections["ta_timeseries"].cursor() as cursor:
-                cursor.execute(
-                    "SELECT _timescaledb_internal.stop_background_workers();"
+    for app, setting in [
+        ("timeseries", settings.TIMESERIES_ENABLED),
+        ("ta_timeseries", settings.TA_TIMESERIES_ENABLED),
+    ]:
+        try:
+            if setting and (
+                f"shared.django_apps.{app}" in settings.INSTALLED_APPS
```
joseph-sentry, line 35: isn't it guaranteed / up to us if the apps are in the INSTALLED_APPS?

joseph-sentry (commented):
qq

apps/worker/migrate_timeseries.py
joseph-sentry: > worker doesn't install timeseries (only ta_timeseries) while API installs both

let's see if we can't change this in a follow-up so that worker installs both

joseph-sentry (approved):

PR #319, opened 2025-07-15
feat: add new continuous aggregates to compute repo wide aggregates

joseph-sentry:
Fixes CCMRG-1376

The 2 PRs below have more context
https://github.com/codecov/umbrella/pull/311
https://github.com/codecov/umbrella/pull/317

With this change we're trying to avoid recomputing the repo-wide top of the page
aggregates. By using continuous aggregates here we can reduce the number of rows
we have to read to finalize the computation of the aggregates by a factor of the
number of tests in the repo, of course at the cost of some more intermediate
storage and asynchronous compute. By doing this we'll reduce the load we're
putting on the database every time a user goes to the dashboard.

We're choosing to use hierarchical continuous aggregates:
https://docs.tigerdata.com/use-timescale/latest/continuous-aggregates/hierarchical-continuous-aggregates/

This is so materializing the daily aggregates is not as expensive, by breaking
down the computation into hourly chunks first, then materializing the daily
aggregates from the hourly aggregates. The goal is to not spend too much time
materializing the daily aggregates. We can gradually do work over the course of
the day to make that faster.

### DIFF
--- a/libs/shared/shared/django_apps/ta_timeseries/migrations/0018_create_repo_summary_caggs.py
+++ b/libs/shared/shared/django_apps/ta_timeseries/migrations/0018_create_repo_summary_caggs.py
@@ -1,0 +1,132 @@
+from django.db import migrations
+
+from shared.django_apps.migration_utils import RiskyRunSQL
+
+
+class Migration(migrations.Migration):
+    atomic = False
+    dependencies = [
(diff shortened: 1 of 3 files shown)


joseph-sentry:
@sentry /review

joseph-sentry:
bugbot run

trent-codecov (approved):

PR #322, opened 2025-07-15
fix(worker): Resolve SQLAlchemy model warning

ElioDiNino:
Ref: https://docs.sqlalchemy.org/en/14/errors.html#error-qzyx

```
WARNING: /app/apps/worker/helpers/log_context.py:89: SAWarning: relationship 'CommitNotification.commit' will copy column commits.id to column commit_notifications.commit_id, which conflicts with relationship(s): 'Commit.notifications' (copies commits.id to commit_notifications.commit_id), 'CommitNotification.commits' (copies commits.id to commit_notifications.commit_id). If this is not the intention, consider if these relationships should be linked with back_populates, or if viewonly=True should be applied to one or more if they are read-only. For the less common case that foreign key constraints are partially overlapping, the orm.foreign() annotation can be used to isolate the columns that should be written towards.   To silence this warning, add the parameter 'overlaps="commits,notifications"' to the 'CommitNotification.commit' relationship. (Background on this error at: https://sqlalche.me/e/14/qzyx)
```


ajay-sentry:
bugbot run

giovanni-guidini (approved):
LGTM, seems to make sense with the ref
(thanks for linking it!)

PR #335, opened 2025-07-18
db: remove default continuous aggregate indexes and create new ones

joseph-sentry:
add migrations to remove the default group indexes create on the new aggregate CAs and add composite indexes on the fields we end up filtering by.


ajay-sentry (approved):
lets do it

PR #341, opened 2025-07-22
chore: `mypy` error and warning fixes

ElioDiNino:
Resolves a number of `mypy` errors and warnings I discovered while working on other changes


apps/codecov-api/services/yaml.py
```diff
@@ -48,6 +48,10 @@ def fetch_commit_yaml(
         yaml_str = async_to_sync(fetch_current_yaml_from_provider_via_reference)(
             commit.commitid, repository_service
         )
+        if yaml_str is None:
+            raise ValueError(
+                f"Yaml not found for commit {commit.commitid} in repository {commit.repository.name}"
```
joseph-sentry, line 53: we should have error messages be static and include dynamic information in the extra dict

apps/codecov-api/upload/views/legacy.py
```diff
@@ -327,8 +327,10 @@ def post(self, request, *args, **kwargs):
...
         ):
-            # if gitlab ci - change domain based by referer
-            build_url = f"{get_config((repository.service, 'url'))}/{owner.username}/{repository.name}/{upload_params.get('build')}"
+            # if gitlab ci - change domain based by referrer
+            build_url: str | None = (
+                f"{get_config((repository.service, 'url'))}/{owner.username}/{repository.name}/{upload_params.get('build')}"
+            )
         else:
             build_url = upload_params.get("build_url")
         queue_params = upload_params.copy()
```
joseph-sentry, line 336: maybe we refactor all this code to end up with something like:

```
build_url: str | None = upload_params.get("build_url")

if repository.service == "gitlab_enterprise" ... :
    build_url = ...

queue_params = ...
```

apps/codecov-api/services/yaml.py
```diff
@@ -83,3 +87,4 @@ def final_commit_yaml(
 def get_yaml_state(yaml: UserYaml) -> YamlStates | None:
     if yaml == get_config("site", default={}):
         return YamlStates.DEFAULT
+    return None
```
ajay-sentry, line 90: this is implicit isn't it?

apps/codecov-api/upload/views/legacy.py
```diff
@@ -347,7 +349,7 @@ def post(self, request, *args, **kwargs):
             # These values below might be different from the initial request parameters, so overwrite them here to ensure they're up-to-date
             "commit": commitid,
             "branch": branch,
-            "pr": pr,
+            "pr": str(pr) if pr else None,
```
ajay-sentry, line 352: did this used to be an int sometimes or smth?

libs/shared/shared/yaml/fetcher.py
```diff
@@ -69,14 +70,16 @@ async def determine_commit_yaml_location(
         )
         if yaml_inside_folder:
             return yaml_inside_folder
+    return None
```
ajay-sentry, line 73: this should be implicit too

libs/shared/shared/yaml/fetcher.py
```diff
@@ -69,14 +70,16 @@ async def determine_commit_yaml_location(
...
 def _search_among_files(
     desired_filenames: Sequence[str], all_files: Sequence[Mapping[str, Any]]
-) -> str:
+) -> str | None:
     for file in all_files:
-        if (
-            file.get("name") in desired_filenames
-            or file.get("path").split("/")[-1] in desired_filenames
+        file_path = file.get("path")
+        if file.get("name") in desired_filenames or (
```
ajay-sentry, line 81: for this one, is this to handle Nones? Or will filePath sometimes be an object or something else?

apps/codecov-api/upload/views/legacy.py
ElioDiNino: Good idea

libs/shared/shared/yaml/fetcher.py
ajay-sentry: If its nones we could maybe just set a default like (file.get("path") or '')

apps/codecov-api/upload/helpers.py
```diff
@@ -348,20 +348,26 @@ def determine_upload_branch_to_use(
...
     # The value of pr can be "true" and we use that info when determining upload branch, however we don't want to save that value to the db
     elif upload_params.get("pr") == "true":
-        return None
+        pass
     else:
-        return upload_params.get("pr")
+        candidate = upload_params.get("pr")
+
+    if candidate and candidate.isdigit():
+        return int(candidate)
```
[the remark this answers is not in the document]
ElioDiNino: Before we were leaving it as a string and then later implicitly casting it to an int when saving it to the database which seemed unsafe, so instead I refactored to return an int which is what we expect its real type to be.

apps/codecov-api/services/yaml.py
ElioDiNino: Yes, but leaving it as implicit causes a bunch of warnings (since it's bad practice) and leads us to miss adding `None` as a possible return value in the type hints (like in this case)

apps/codecov-api/upload/views/legacy.py
ElioDiNino: It used to be `str | None` but with the function change above it now returns `int | None`. For consistency with how the task calling is working I casted it back to a string (although I couldn't find a downstream use of it)

libs/shared/shared/yaml/fetcher.py
ElioDiNino: See comment about implicit above

libs/shared/shared/yaml/fetcher.py
ElioDiNino: > for this one, is this to handle Nones? Or will filePath sometimes be an object or something else?

A mix of both. According to the typing the result could be `Any` (although it's possible in practice we only see strings)

> If its nones we could maybe just set a default like (file.get("path") or '')

Yeah but we don't want to run the split and `in desired_filenames` in that case imo and we still need to check it's a string due to the above

apps/codecov-api/services/yaml.py
ElioDiNino: To each their own I guess, but `mypy` said this (prior to my changes):
image

libs/shared/shared/yaml/fetcher.py
ajay-sentry: If it could be values other than strings then the current implementation makes sense ✅

apps/codecov-api/services/yaml.py
ajay-sentry: Did a little more digging and apparently in PEP8 it was called out that standard python syntax should explicitly return None's so I stand corrected lol
https://peps.python.org/pep-0008/#programming-recommendations

The line starting with "Be consistent in return statements...."

Weird bc the same thing isn't explicit in TypeScript

ElioDiNino: Good to know haha

> Weird bc the same thing isn't explicit in TypeScript
The joys of the JS/TS world lol

PR #345, opened 2025-07-23
fix: Properly install Django stubs with `mypy`

ElioDiNino:


apps/codecov-api/mypy.ini
```diff
@@ -1,11 +1,18 @@
...
 follow_imports = silent
 warn_no_return = False
 
 [mypy-*.tests.*]
-disallow_untyped_defs = False
\ No newline at end of file
+disallow_untyped_defs = False
+
+[mypy.plugins.django-stubs]
+django_settings_module = "codecov.settings_enterprise"
```
ajay-sentry, line 18: why enterprise settings vs. another confidence level?

apps/codecov-api/mypy.ini
```diff
@@ -1,11 +1,18 @@
 # Global options:
 
 [mypy]
+plugins =
+    mypy_django_plugin.main,
```
ajay-sentry, line 5: Why are we adding these plugins?

.vscode/settings.json
```diff
@@ -12,6 +12,7 @@
     "apps/worker",
     "libs/shared",
   ],
+  "mypy-type-checker.cwd": "${nearestConfig}",
```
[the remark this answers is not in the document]
ElioDiNino: Yes, because we don't have a root-level config (the default), so the `mypy` extension won't use the configuration files we currently have for each app

apps/codecov-api/mypy.ini
ElioDiNino: Since it's the settings file with the most configurations specified (pretty sure every other settings file has a strict subset of its options but I haven't fully verified that). This matters because with the `mypy` plugins I added, you get a warning if you reference a Django setting that isn't defined.

apps/codecov-api/mypy.ini
ElioDiNino: Since Django and Django rest framework don't come with typing, these plugins (which we already had installed but not configured), add the necessary typing information for `mypy`.

ElioDiNino:
> Can we add some context in the PR description?

Done!

.vscode/settings.json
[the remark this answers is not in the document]
ajay-sentry: oooo okay nice, nearestConfig is a special var then?

PR #346, opened 2025-07-23
fix: aggregate test results from TA hypertable if data's not in CAs

joseph-sentry:
Fixes CCMRG-1395

the continuous aggregates only rollup raw testrun data at a branch level
granularity for certain branches. For branches that aren't being rolled up we
can't fetch the data from CAs, instead we have to aggrgate directly from the
testrun hypertable.

We don't have to compute test results aggregates or flake aggregates yet, as
that adds too much complexity for now, and we'll probably test out a new CA
where we compute rollups for all branches before we build that.


trent-codecov (approved):

.vscode/extensions.json
ElioDiNino, line 1: Why are we removing `mypy` if we have `mypy` configured for each project? Why don't we just add this as an additional tool

PR #361, opened 2025-07-29
fix: wrap new TA impl in try except for cache rollups

joseph-sentry:
otherwise if the new impl fails, we'll crash and never get to the old impl

### DIFF
--- a/apps/worker/tasks/cache_test_rollups.py
+++ b/apps/worker/tasks/cache_test_rollups.py
@@ -2,6 +2,7 @@ import datetime as dt
 from typing import Literal
 
 import polars as pl
+import sentry_sdk
 from django.db import connections
 from redis.exceptions import LockError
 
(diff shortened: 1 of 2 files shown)


apps/worker/tasks/cache_test_rollups.py
```diff
@@ -125,10 +126,15 @@ def run_impl(
...
-                        return {"success": True}
+                    try:
+                        cache_rollups(repo_id, branch)
+                        cache_rollups(repo_id, None)
+                        if impl_type == "new":
+                            return {"success": True}
+                    except Exception as e:
+                        sentry_sdk.capture_exception(e)
+                        if impl_type == "new":
+                            return {"success": False}
```
ElioDiNino, line 137: Confirming that we still want to proceed with the rest of the function in the case of an exception if `impl_type` is not "new"?

apps/worker/tasks/process_flakes.py
```diff
@@ -69,9 +70,14 @@ def run_impl(
...
-            if impl_type == "new":
-                return {"successful": True}
+            try:
+                process_flakes_for_repo(repo_id)
+                if impl_type == "new":
+                    return {"successful": True}
+            except Exception as e:
+                sentry_sdk.capture_exception(e)
+                if impl_type == "new":
+                    return {"successful": False}
```
ElioDiNino, line 80: Same as above comment

apps/worker/tasks/cache_test_rollups.py
joseph-sentry: yeah we want to continue in the case of "both" but we still want to capture the exception in Sentry

PR #367, opened 2025-07-31
fix: show up to date TA data

joseph-sentry:
I'm changing the bounds on the TA data to show the most up to date data we can show. Given we've enabled real time aggregates


joseph-sentry:
@sentry review

PR #372, opened 2025-08-01
fix: temporarily default to default branch data until we update the FE

joseph-sentry:
gazebo doesn't explicitly pass the branch to flake aggregates or test results aggregates so having the default be all branches is confusing for users because the table data is showing default branch data but the metrics section is showing all branches data


ElioDiNino (approved):
LGTM

PR #373, opened 2025-08-01
Tidy up TA timescale rollup and old CAs

joseph-sentry:
deps on #354

- remove TA GQL read from rollup function and tests
- remove code that did TA GCS rollups from timescale
  - since the API is reading directly from timescale now, we don't need to do this anymore, so i'm deleting the related code
- remove old CAs
- update retention policy on new CAs
  - we don't need to store data for all time for the hourly CAs just more recent data, and we want to limit the retention on data in the new CAs


michelletran-codecov (approved):
LGTM. Just make sure to test this before shipping.

PR #375, opened 2025-08-04
chore: search repo by repoid in admin list

giovanni-guidini:
Check [this ticket](https://linear.app/getsentry/issue/CCMRG-1449/request-to-delete-specified-repositories#comment-78ca9d35) for full details, but the list of repos in the admin view doesn't list them by repoID, rather it's by owner username.

This is annoying - and might get in the way depending on what you are trying to do. So these changes, fully based on Django admin docs, with tests by Cursor, also let us search by repod ID.

### Legal Boilerplate


ajay-sentry (approved):
lgtm

PR #386, opened 2025-08-07
feat(api): Flatten directories with single child directory for tree v…

calvin-codecov:
Closes https://linear.app/getsentry/issue/CCMRG-1219/snowflake-flatten-file-paths-in-the-file-viewer-component

This PR modifies how the "tree" viewer shows directories that have a single directory as its only child by flattening. This scenario occurs often in java/kotlin/go projects and while our "list" viewer shows full file paths, it is not the most efficient way to navigate to files.

SOLUTION: If a directory has a single child that is a directory, we tack on the child to the parent and link them together. This is similar to how GitHub shows their tree directories and gives users quick navigation with clarity. (This is slightly different from the proposed solution in the ticket).

Flatten example:
```
src/a/file1.py
src/a/b/c/d/file2.py
```
Clicking into directory `src/a` will show a link to file1.py and a flattened link to `/b/c/d` as `b` and `c` only have one child that is directory. 

Note:
I originally implemented a solution that started flattening at a directory that was an only child and only had one child, directory or file, as its child but I decided that this was not as useful as its criteria was very narrow. For example, for the above input, even though `/b` only has one child, because it has a sibling in file1.py, the flattening would not start until inside of `/b` for `/c/d/file2.py`. If it were to start earlier inside `/a`, it could have been potentially confusing for the user seeing a link to a file and a link to file with flatten dirs in between.
Ex:
```
Viewing /src/a
- file.py
- /b/c/d/file2.py
```

https://github.com/user-attachments/assets/52fe265b-a5f7-440b-ace9-4a374cd752f3

### DIFF
--- a/apps/codecov-api/services/path.py
+++ b/apps/codecov-api/services/path.py
@@ -243,36 +243,25 @@ class ReportPaths:
             grouped[path.basename].append(path)
 
         results: list[File | Dir] = []
-        # check if current level has multiple groups to allow for flattening
-        has_siblings = len(grouped) > 1
 
         for basename, paths in grouped.items():
-            if len(paths) == 1:
-                if paths[0].is_file:
-                    path = paths[0]
-                    results.append(
-                        File(full_path=path.full_path, totals=self._totals(path))
-                    )
-                else:
-                    # continue on if there is only one child and it is a directory
-                    children = self._single_directory_recursive(
-                        PrefixedPath(full_path=path.full_path, prefix=basename)
-                        for path in paths
-                    )
-
-                    # we don't flatten directories with a single child if that parent
-                    # directory has siblings
-                    if has_siblings:
-                        results.append(Dir(full_path=basename, children=children))
-                    else:
-                        # we flatten directories with a single child when they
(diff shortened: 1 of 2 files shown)


apps/codecov-api/services/path.py
```diff
@@ -26,7 +26,15 @@ class PathNode:
 
     @property
     def name(self) -> str:
-        return self.full_path.split("/")[-1]
+        if hasattr(self, "children"):  # this is a Dir
+            max_levels = getattr(self, "max_directory_level", 1)
```
ajay-sentry, line 30: will this ever not be 1?

apps/codecov-api/services/path.py
```diff
@@ -70,6 +78,7 @@ class Dir(PathNode):
 
     full_path: str
     children: list[File | Dir]
+    max_directory_level: int = 1
```
ajay-sentry, line 81: just for the future, any chance we can add some comments to these fields while the context you have is still fresh?

ajay-sentry (approved):
lgtm! thanks for walkthrough

PR #388, opened 2025-08-08
fix: was_squash_via_merge_commit logic

joseph-sentry:
Check if the merge commit sha is equal to the sha of the head of the PR because in that case we should also move the commits on the PR to the base branch of the PR.


michelletran-codecov (commented):
> Check if the merge commit sha is equal to the sha of the head of the PR because in that case we should also move the commits on the PR to the base branch of the PR.

What's the context for this change? The rebased changes will have a different SHA, so moving these commits to the base branch will cause a mismatch with commit SHA and the actual commit on the main branch.

joseph-sentry:
> What's the context for this change? The rebased changes will have a different SHA, so moving these commits to the base branch will cause a mismatch with commit SHA and the actual commit on the main branch.

There's a case where the experiment was failing because the head sha and merge commit sha were equivalent. The previous logic in the experiment would count that as a squash, because it was only checking the # of parents, but the previous logic was correctly detecting that as not a squash merge.

michelletran-codecov (approved):
Seems low risk given the feature flag. 👍

PR #392, opened 2025-08-08
ref: github webhooks

joseph-sentry:
get rid of status_set_pending task since that's only used in a part of the github webhook handler that never gets called so its essentially dead code

get rid of unused logs in github webhook handler

add a log to check if any users are using the ci skip feature


apps/codecov-api/webhook_handlers/views/github.py
```diff
@@ -53,6 +48,9 @@ class GithubWebhookHandler(APIView):
 
     service_name = "github"
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
ElioDiNino, line 52: This is redundant since the super's init function will be called when the subclass doesn't define its own init function

PR #393, opened 2025-08-08
feat: add endpoint for ingesting webhooks forwarded from Sentry

joseph-sentry:
depends on #392 

add dictionary for routing request to a custom Sentry webhook handler
or the regular github handlers. this is because we don't want to process
all webhooks sent to us by sentry, and we also need custom processing
for other events, such as "pull_request"

use jwt auth permission class, since we'll be receiving these requests
via the codecov client in Sentry

emit logs and sentry exceptions on any exception during handling for
observability


apps/codecov-api/webhook_handlers/views/sentry.py
```diff
@@ -0,0 +1,79 @@
...
+        handler = SENTRY_EVENT_HANDLERS.get(action)
+        if handler is None:
+            return Response({"status": "ok"})
+
+        try:
+            github_webhook_handler.event = action
+            handler(request)
+        except Exception as e:
+            sentry_sdk.capture_exception(e)
+            log.error(
```
ElioDiNino, line 73: `log.exception(...)` will capture the trace and full error details (no need for `"error": str(e)`)

apps/codecov-api/webhook_handlers/views/sentry.py
```diff
@@ -0,0 +1,79 @@
...
+
+        sentry_webhook.labels(event=action).inc()
+        log.info(
+            f"Received {action} webhook",
+            extra={"payload": request.data},
+        )
+
+        handler = SENTRY_EVENT_HANDLERS.get(action)
+        if handler is None:
+            return Response({"status": "ok"})
```
ElioDiNino, line 66: Should we not log or indicate an error here since we want to know if we ever have trouble processing webhooks from Sentry? This is assuming that within Sentry we are picking out the webhooks being send (vs firehosing everything)

ElioDiNino (commented):
Generally looks good, just a few questions and comments

apps/codecov-api/webhook_handlers/views/sentry.py
```diff
@@ -0,0 +1,79 @@
...
+SENTRY_EVENT_HANDLERS = {
+    "repository": github_webhook_handler.repository,
+    "delete": github_webhook_handler.delete,
+    "member": github_webhook_handler.member,
+    "organization": github_webhook_handler.organization,
+    "installation": github_webhook_handler.installation,
+    "installation_repositories": github_webhook_handler.installation_repositories,
+    "public": github_webhook_handler.public,
+    "pull_request": handle_pull_request,
+}
```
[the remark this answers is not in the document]
ElioDiNino: Also the "status" event

PR #396, opened 2025-08-11
fix(api): Make all config values use standard config function

ElioDiNino:
This allows configuration via both environment variables or via the config file instead of just the former


joseph-sentry (approved):
can you paste the relevant snippet of the get_config function that makes it so the env var settings override all other possible settings so if those env vars are set the behaviour is unchanged?

there is the case where behaviour will change if the values are specified in the config but not in env vars so let's maybe double check that before merging just in case

PR #398, opened 2025-08-11
fix: update webhook github app installation handling

joseph-sentry:
We're essentially treating Sentry apps like the custom ones that are defined through the database, by including the app id and pem path when creating the DB object in the webhook handler.

we're trusting that the logic that was written for supporting those custom apps defined through the DB works properly and with that assumption, this approach should work.

### DIFF
--- a/apps/codecov-api/webhook_handlers/tests/test_github.py
+++ b/apps/codecov-api/webhook_handlers/tests/test_github.py
@@ -5,6 +5,7 @@ from hashlib import sha1, sha256
 from unittest.mock import call, patch
 
 import pytest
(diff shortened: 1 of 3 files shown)


joseph-sentry:
i decided to minimize the changes to the point where it's just enough to have the sentry app installations work, i didn't consider that we'd have to update the settings in all the apps. In the future it would be nice to have those defined as a var in shared or something though.

ElioDiNino (commented):
Can we add a quick test for this please. Looks good otherwise

ElioDiNino (approved):

PR #400, opened 2025-08-13
fix(billing): check admin is in org before sending email

joseph-sentry:
- check an admin is part of an org before sending an email
- filter the admin owner ID if they're not in the org
- emit a warning message for observing how common this logical error is

- TODO: might need to queue up a refresh task instead of just removing the invalid owner from the admin list


ajay-sentry:
What's prompting this change? Are stale admins something we should be fixing upstream rather than relying on a separate process to do the pruning?

PR #407, opened 2025-08-14
fix: update sync_teams and ref stripe webhooks

joseph-sentry:


apps/worker/helpers/admins.py
```diff
@@ -0,0 +1,32 @@
...
+from database.models import Owner
+
+log = logging.getLogger(__name__)
+
+
+def update_single_owner_admins(db_session: Session, owner: Owner) -> None:
+    if not owner.admins:
+        return
+
+    admin_ids = owner.admins if owner.admins is not None else []
```
ElioDiNino, line 14: Due to the `if` check above, isn't the `owner.admins is not None` check unnecessary?

libs/shared/shared/django_apps/codecov_auth/models.py
```diff
@@ -696,6 +696,30 @@ def remove_admin(self, user):
                 pass
         self.save()
 
+    def update_admins(self) -> None:
+        if not self.admins:
+            return
+
+        admin_ids = self.admins if self.admins is not None else []
```
ElioDiNino, line 703: Same comment as above

ElioDiNino (commented):
Two small comments. Also wondering if there's any way to reduce to duplicated code between worker and shared (which gets used for API)

PR #408, opened 2025-08-15
feat: detect flakes task implementation

joseph-sentry:
we implement the detect flakes task that uses the new upload object to choose
which testruns to fetch from timescale to process.

it takes an indefinite redis lock so it's the only detect flakes task running
for a given repo, and it runs the DB updates for a given upload in a transaction
so that if we get a celery task timeout, we don't leave an upload half-processed


libs/shared/shared/celery_config.py
```diff
@@ -73,6 +73,12 @@
     f"app.tasks.{TaskConfigGroup.test_results.value}.SyncTestResultsTask"
 )
 
+ingest_testruns_task_name = (
+    f"app.tasks.{TaskConfigGroup.test_results.value}.IngestTestruns"
+)
+
```
ElioDiNino, line 79: Assuming this is included from #403?

apps/worker/tasks/detect_flakes.py
```diff
@@ -0,0 +1,90 @@
...
+                                created = Flake.objects.filter(
+                                    repoid=repo_id, test_id__in=new_test_ids
+                                )
+                                for flake in created:
+                                    curr_flakes[bytes(flake.test_id)] = flake
+                            upload.delete()
+            return True
+    except LockError:
+        log.warning("Failed to acquire lock for repo %s", repo_id)
+        return False
```
ElioDiNino, line 74: Might be worth adding a comment here explaining that it's safe to exit since another detect flakes task will pick up the necessary work

ElioDiNino (approved):

PR #422, opened 2025-08-20
feat(worker): New TA notifier task

ElioDiNino:
Depends on #418 and #421 

* Resolved some pytest warnings due to class names
* Added Sentry app name setting to worker prod and shared testing
* Added the new TA notifier task

### DIFF
--- a/apps/worker/tasks/test_analytics_notifier.py
+++ b/apps/worker/tasks/test_analytics_notifier.py
@@ -1,0 +1,343 @@
+import logging
+from contextlib import contextmanager
+from typing import cast
+
+from asgiref.sync import async_to_sync
+from django.conf import settings
+from redis import Redis
+from sqlalchemy.orm import Session
+
+from app import celery_app
+from database.enums import ReportType
+from helpers.notifier import NotifierResult
+from helpers.string import shorten_file_paths
+from services.lock_manager import LockManager, LockRetry, LockType
+from services.repository import (
+    fetch_pull_request_information,
+    get_repo_provider_service,
+)
+from services.test_analytics.ta_metrics import (
+    read_failures_summary,
+    read_tests_totals_summary,
+)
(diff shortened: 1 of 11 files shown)


apps/worker/tasks/test_analytics_notifier.py
```diff
@@ -0,0 +1,331 @@
...
+                "Acquired fencing token, retrying for debounce period",
+                extra=self.extra_dict,
+            )
+            self.retry(countdown=DEBOUNCE_PERIOD_SECONDS, kwargs=kwargs)
+
+        # At this point we have a fencing token, but want to check if another
+        # notification task has incremented it, indicating a newer notification
+        # task is in progress and can take over
+        assert fencing_token, "Fencing token not acquired"
+        with self._notification_lock(lock_manager):
```
joseph-sentry, line 118: i'd write a comment here that mentions that the lock here is not required for correctness, but the one at the end is.

the lock + fencing token check here guards against the situation where we do duplicate work preparing the notification (reads against the db), but this is just saving our db from unneeded read load. we could get rid of it and still be correct

apps/worker/tasks/test_analytics_notifier.py
```diff
@@ -0,0 +1,331 @@
...
+            lock_timeout=max(80, self.hard_time_limit_task),
+            redis_connection=redis_client,
+        )
+
+        if fencing_token is None:
+            with self._notification_lock(lock_manager):
+                # Acquire a fencing token to deduplicate notifications
+                fencing_token = cast(
+                    int,
+                    redis_client.incr(
```
joseph-sentry, line 98: i think we need a TTL on this? looks like we'll need to use a redis multi statement with expire (let's say 24 hours?)

the choice to use lt instead eq in check_fencing_token below supports this:
if we have an expiry between the incr and the expire, that task will still go through because it'll be strictly above the val of current_token

PR #437, opened 2025-08-26
fix(api): Configurable proxy depth

ElioDiNino:
This allows us to resolve a security vulnerability where an attacker can initiate their request with an arbitrary IP address in the  X-Forwarded-For header and we trust it implicitly.

Resolves (once configured in Helm) https://linear.app/getsentry/issue/CCMRG-1328/ip-address-spoofing-via-x-forwarded-for-header-manipulation


apps/codecov-api/codecov/settings_base.py
```diff
@@ -280,6 +280,11 @@
 
 
 # Session configuration
+
+# This controls how the X-Forwarded-For header is processed to get client IP addresses
+# Defaults to trusting all proxies and taking the first IP in the list
+TRUSTED_PROXY_DEPTH = get_config("setup", "http", "trusted_proxy_depth", default=0)
```
mdtro, line 286: I find this setting a bit confusing. A negative is later applied in the `get_client_ip_address`. 

If I wanted the second value in the list, I would initially set this to `1` -- counting from the left, indexed at `0` as indicated by the default value. 

This would be incorrect in GCP with a load balancer involved. It would return the last value in the list because of the applied negative.

Additional comments might help here indicating we count from the right, or instead just take the target index value directly?

ElioDiNino: > I find this setting a bit confusing. A negative is later applied in the get_client_ip_address.

That's fair, but the idea is that you do not need to look at the underlying code but instead configure what the setting is asking for: how many trusted proxies do you have in front of the application

> If I wanted the second value in the list, I would initially set this to 1 -- counting from the left, indexed at 0 as indicated by the default value.

`X-Forwarded-For` should always be parsed from the right, not the left. This can be seen in the MDN docs: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For#trusted_proxy_count

> This would be incorrect in GCP with a load balancer involved. It would return the last value in the list because of the applied negative.

If on GCP and not using any additional proxies then the trusted depth would just be 2 as outlined in [their docs](https://cloud.google.com/load-balancing/docs/https#x-forwarded-for_header)

> Additional comments might help here indicating we count from the right, or instead just take the target index value directly?

How does this adjusted comment sound to you?

```python
# Controls how the X-Forwarded-For header is processed to get client IP addresses.
# Set to how many trusted proxies you have in front of the application (must be
# a positive integer). Defaults to trusting all proxies (this is unsafe unless
# your ingress proxy strips X-Forwarded-For headers).
```

michelletran-codecov (approved):
Approving this for now, as it's solving the current incorrect behaviour for the app 😶 .

I'm kind of unsure that we should be configuring the network behaviour as a config in the app. We should probably be stripping the X-Forwarded-For header at the network ingress level. I can talk to @jason-ford-codecov to see if we can change this at a network infra level. In the meantime, we can just merge this change for now and remove the config later.

PR #441, opened 2025-08-29
fix: django state

joseph-sentry:
i created an index manually but forgot to modify the django state so  this commit is making sure django is aware of the index having been  Created


ElioDiNino (approved):
Can confirm this resolves the warning

PR #442, opened 2025-08-29
feat: update migration_utils

joseph-sentry:
we create a function to allow to run a timescale index creation with 
transaction per chunk and modify the django state at the same time

we also create functions for removing the retention policy on a CA and 
create a TTL per row retention policy for a hypertable and for a CA


ElioDiNino (approved):
See comments but generally LGTM

adrianviquez (approved):

PR #444, opened 2025-08-29
ref: add retention policy, remove old CAs, stop cache rollups in new TA

joseph-sentry:
we remove the old CAs that are no longer being used in the GQL api since
we’re using the new ones exclusively

the hourly CAs are just used as a “bridge” to the daily CAs so we do 
some intermediary deduplication within the hour, just so we amortize the
cost of aggregation a bit more throughout the day, so we don’t have to 
retain them for long

we don’t need to cache rollups in the new pipeline anymore since the API
is consuming from timescale directly

So we’re removing all code related to caching rollups in the timescale 
implementation of the TA pipeline


ElioDiNino (approved):
Lovely cleanup 🧹

adrianviquez (approved):

PR #447, opened 2025-09-02
chore: temporarily remove error messages from regular PR comments

joseph-sentry:
we wanted to include upload errors in the PR comments but they’re quite  noisy in their current implementation and we want to make sure to not  accidentally clobber the PR comment when we rollout the TA timescale  implementation to more users so we’re going to temporarily remove the  upload errors from all non-error comments: if the comment is just  notifying the error, then we will included, if there are _some_ valid  results, then we will not include it at all.


michelletran-codecov (commented):
Generally LGTM. Just nitpick and question above.

michelletran-codecov (approved):

PR #452, opened 2025-09-03
fix: add metrics for get test results timescale

joseph-sentry:
we need to run the `connection.edges` line so that we populate the property under the metric so we have an actual idea of how long it took to run that query, otherwise the query would probably run somewhere in the GQL library code

### DIFF
--- a/apps/codecov-api/graphql_api/types/test_analytics/test_analytics.py
+++ b/apps/codecov-api/graphql_api/types/test_analytics/test_analytics.py
@@ -27,6 +27,7 @@ from graphql_api.types.test_results_aggregates.test_results_aggregates import (
 )
 from rollouts import READ_NEW_TA
 from shared.django_apps.core.models import Repository
+from shared.metrics import Histogram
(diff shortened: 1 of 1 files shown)


apps/codecov-api/graphql_api/types/test_analytics/test_analytics.py
```diff
@@ -362,52 +368,56 @@ async def resolve_test_results(
...
-            else TestResultsOrderingParameter.AVG_DURATION
-        )
-        ordering_direction = (
-            ordering.get("direction", OrderingDirection.DESC)
-            if ordering
-            else OrderingDirection.DESC
-        )
+        with test_results_histogram.time():
+            measurement_interval = (
+                filters.get("interval", MeasurementInterval.INTERVAL_30_DAY)
```
adrianviquez, line 373: Nit: the double mention of `MeasurementInterval.INTERVAL_30_DAY` feels a bit odd. I'd do

`filters.get("interval") if filters else MeasurementInterval.INTERVAL_30_DAY`

If you need to check `filters`

PR #453, opened 2025-09-03
feat: restore TA GQL resolvers to defaulting to all branches

joseph-sentry:
we needed to update the UI to always pass the branch param, even when  the branch is the default branch, and now that that is done, we can  safely set the default here to be “all branches"

also refs the measurement interval code in test analytics

### DIFF
--- a/apps/codecov-api/graphql_api/types/test_analytics/test_analytics.py
+++ b/apps/codecov-api/graphql_api/types/test_analytics/test_analytics.py
@@ -175,7 +175,7 @@ async def resolve_test_results_aggregates(
     start_date = end_date - timedelta(days=measurement_interval.value)
     return await sync_to_async(get_test_results_aggregates)(
         repoid=repository.repoid,
-        branch=repository.branch,
+        branch=branch,
         start_date=start_date,
         end_date=end_date,
     )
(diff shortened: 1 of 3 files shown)


apps/codecov-api/graphql_api/types/test_analytics/test_analytics.py
```diff
@@ -175,7 +175,7 @@ async def resolve_test_results_aggregates(
     start_date = end_date - timedelta(days=measurement_interval.value)
     return await sync_to_async(get_test_results_aggregates)(
         repoid=repository.repoid,
-        branch=repository.branch,
+        branch=branch,
```
giovanni-guidini, line 178: This change addresses a bug where the branch parameter wasn't being passed correctly to the underlying functions. Previously, it always used `repository.branch` even when a specific branch was provided to the resolver. The fix correctly passes the branch parameter from the GraphQL query to the data retrieval functions.

This is a good fix as it allows users to specify which branch they want to get analytics for, rather than always defaulting to the repository's default branch.

apps/codecov-api/graphql_api/types/test_analytics/test_analytics.py
```diff
@@ -194,7 +194,7 @@ async def resolve_flake_aggregates(
     start_date = end_date - timedelta(days=measurement_interval.value)
     return await sync_to_async(get_flake_aggregates)(
         repoid=repository.repoid,
-        branch=repository.branch,
+        branch=branch,
```
giovanni-guidini, line 197: Similar to the previous fix, this change ensures the branch parameter from the GraphQL query is correctly passed to the flake aggregates function. This is important for consistency in the API behavior.

PR #455, opened 2025-09-03
fix: Update plan when calling account link endpoint

ajay-sentry:
This PR updates the account link code to use the new sentry merge plan rather than the default plan since that's the one we'll be using for now to determine which users are sentry account users.

Relates to https://linear.app/getsentry/issue/CCMRG-1331/sentry-hook-into-sentry-github-app-install-to-call-codecov

### Legal Boilerplate


michelletran-codecov (approved):
LGTM!

PR #458, opened 2025-09-04
feat: update TA notify to use sentry app depending on owner account

joseph-sentry:
I made some change to the webhooks where i check the account of the  owner to decide which apps ingests the webhook and i’m using the  same approach here to determine which app will make the comment in the  TA notifier


michelletran-codecov (commented):
Generally LGTM. But weren't we going to use a different message for prevent? If we're using the same message, then this should be fine.

PR #465, opened 2025-09-09
ref: tidy up old TA code that is no longer used

joseph-sentry:
given the migration to timescale for storing TA data, all this code being removed is obsolete

- remove usage of test instances models in worker
- remove usage of cache files in API
- remove old test analytics models


michelletran-codecov (commented):
I would be more confident if we pushed out the update to the API before removing model. How much traffic are we expecting from the public API for test analytics?

joseph-sentry:
> I would be more confident if we pushed out the update to the API before removing model. How much traffic are we expecting from the public API for test analytics?

not much at all

michelletran-codecov (approved):
OK, approving this because risk feels low given the low traffic on TA public APIs.

PR #466, opened 2025-09-09
feat: Account linking/unlinking updates

ajay-sentry:
This PR does a couple things:

- Updates account linking logic to throw an error for already linked accounts on github orgs
- Bypasses account creating for orgs with existing accounts
- Reactivates accounts and updates their sentry org details if we find an existing inactive account
- Creates a new "unlink" endpoint which does the account deactivation in the event a sentry org removes the prevent add-on.

Still need to fix/add tests but wanted to get eyes on the approach first for any holes I might've missed.

### Legal Boilerplate

### DIFF
--- a/apps/codecov-api/api/sentry/views.py
+++ b/apps/codecov-api/api/sentry/views.py
@@ -1,7 +1,7 @@
 import logging
 
 from django.conf import settings
-from rest_framework import serializers
+from rest_framework import serializers, status
 from rest_framework.decorators import api_view, permission_classes
 from rest_framework.response import Response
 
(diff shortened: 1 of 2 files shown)


apps/codecov-api/api/sentry/views.py
```diff
@@ -99,3 +144,24 @@ def account_link(request, *args, **kwargs):
...
+    sentry_org_id = serializer.validated_data["sentry_org_id"]
+
+    try:
+        account = Account.objects.get(sentry_org_id=sentry_org_id)
+    except Account.DoesNotExist:
+        return Response(
+            {"message": "Account not found"}, status=status.HTTP_404_NOT_FOUND
+        )
+
+    account.is_active = False
```
joseph-sentry, line 164: given the logic in the rest of the code (webhook, github integration stuff) can we also set the sentry_org_id to be null?

i guess more generally, what is the advantage of discerning between reactivation and activation for the first time

joseph-sentry (approved):
quick question just to clarify the approach (i don't mind updating the webhook/github installation logic to take this into account)

PR #473, opened 2025-09-11
feat: use correct app when user has sentry account linked

joseph-sentry:
When an owner has a sentry_org_id linked through their account,
always use the sentry app installation regardless of the
installation_name parameter passed in.

### DIFF
--- a/libs/shared/shared/bots/github_apps.py
+++ b/libs/shared/shared/bots/github_apps.py
@@ -2,6 +2,8 @@ import logging
 import random
 from datetime import UTC, datetime
 
+from django.conf import settings
+
 from shared.bots.exceptions import NoConfiguredAppsAvailable, RequestedGithubAppNotFound
 from shared.bots.types import TokenWithOwner
 from shared.django_apps.codecov_auth.models import (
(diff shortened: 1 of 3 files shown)


libs/shared/shared/bots/github_apps.py
```diff
@@ -261,15 +263,21 @@ def get_github_app_info_for_owner(
...
+            overridden and the sentry app will be used instead. If None and no sentry_org_id exists,
+            the installation name will be determined automatically.
 
     Returns:
         (ordered) List[GithubInstallationInfo]: where index 0 is the main app and the others are fallback options
 
     Raises:
         NoConfiguredAppsAvailable: Owner has app installations available, but they are all currently rate limited.
     """
+    if owner.account and owner.account.sentry_org_id is not None:
```
giovanni-guidini, line 278: The condition `owner.account and owner.account.sentry_org_id is not None` could lead to unexpected behavior if `owner.account` is None. Consider more explicit null-checking to improve readability and maintainability.

Also, there's no documentation on why a sentry_org_id being present should override the installation_name with the sentry app name. This relationship isn't immediately obvious and should be documented in the function's docstring.
```suggestion
    # If owner has a Sentry organization linked, always use the Sentry GitHub app
    if owner.account is not None and owner.account.sentry_org_id is not None:
        installation_name = settings.GITHUB_SENTRY_APP_NAME
```

PR #476, opened 2025-09-12
feat: update conditions for showing error comment and add passing comment

joseph-sentry:
i'm changing the TA behaviour for showing the error comment and adding a passing comment.

we only want to show the error comment if there are no relevant results to show

if we have no failures, but other results, we want to show a passing comment

### DIFF
--- a/apps/worker/services/test_analytics/ta_finish_upload.py
+++ b/apps/worker/services/test_analytics/ta_finish_upload.py
@@ -161,7 +161,9 @@ def ta_finish_upload(
     with read_tests_totals_summary.labels(impl="new").time():
         summary = get_pr_comment_agg(repoid, commitid)
 
-    if not summary["failed"] and error is None:
+    num_testruns = sum(summary.values())
+
+    if num_testruns == 0 and error is None:
         log.info(
             "No failures and no error so not posting comment but still queueing notify",
             extra=extra,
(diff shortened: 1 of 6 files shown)


apps/worker/services/test_analytics/ta_finish_upload.py
```diff
@@ -209,16 +209,24 @@ def ta_finish_upload(
             "queue_notify": False,
         }
 
-    if summary["failed"] == 0:
-        # no failures, only error
-        log.info("No failures, posting error comment", extra=extra)
+    if all(value == 0 for value in summary.values()):
```
ajay-sentry, line 212: can we avoid doing the summary.values() loop 3 times on 164,212 and 222 by just setting values to a const somewhere like "has_value" = sum(summary.values())

apps/worker/services/test_analytics/ta_processor.py
```diff
@@ -30,14 +32,7 @@ def ta_processor(
...
 ) -> bool:
-    log.info(
-        "Processing single TA argument",
-        extra={
-            "upload_id": argument.get("upload_id"),
-            "repoid": repoid,
-            "commitid": commitid,
-        },
-    )
+    log.info("Processing single TA argument")
```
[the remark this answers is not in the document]
joseph-sentry: those arguments are included in the log context already, no need to have them as well

ajay-sentry (approved):
astounding

apps/worker/services/test_analytics/tests/test_ta_processor.py
```diff
@@ -269,3 +282,205 @@ def test_ta_processor_success_keep_archive(mock_storage, sample_test_json_path):
     assert testrun_db.flags == [flag.flag_name]
 
     assert mock_storage.read_file("archive", "path/to/valid.json") is not None
+
+
+@pytest.mark.django_db
```
giovanni-guidini, line 287: The test file has grown significantly after merging content from the deleted test_ta_processing.py. Consider creating separate test classes to better organize the different test scenarios and improve readability.
```suggestion
# Tests for functionality moved from ta_processing.py
class TestTAProcessingFunctionality:
    @pytest.mark.django_db
    def test_handle_file_not_found(self):
        upload = UploadFactory()

        handle_file_not_found(upload)

        assert upload.state == "processed"

        error = UploadError.objects.filter(report_session=upload).first()
        assert error is not None
        assert error.error_code == "file_not_in_storage"
```

PR #488, opened 2025-09-15
fix: sentry github app installation logic

joseph-sentry:
the code was previously not setting the name correctly. This isn't the greatest
approach since we're handling the sentry github app case specifically and
ideally we'd have a "system" for discerning between "default apps" and
"custom apps" but for now this hard coded handling will have to do.

This commit also updates the logic to no longer set the pem path (because it's
no longer needed) and to make that possible without breaking the existing
behaviour we have to update the definition of the is_configured method on the
github app installation model so it considers a list of default app IDs instead
of just the one.


ajay-sentry (approved):
sure

PR #492, opened 2025-09-15
fix: don't check TA if TA is disabled in notify

joseph-sentry:
we need to make sure we don't try getting data from the TA database is TA is not enabled for the entire codecov instance, which I wasn't doing.

I also do some light refactoring of the source code and the tests.
entry's choice of terms.


drazisil-codecov (approved):
Looks good if the tests pass. :)

PR #496, opened 2025-09-16
fix: pem path overriding and ref get_pem

joseph-sentry:
the code that was here before was confusing and made some false assumptions, so we're refactoring and fixing it.

1.  get_pem now takes in the app ID, service and pem_path of the github app install object, instead of pem_path and pem_name because that was confusing
2.  get_pem_overrides now returns the path of the config value that contains the file path to the PEM

basically the get_pem_overrides function was returning the wrong values and the code was hard to reason about so i refactored the get_pem function so the logic is more flat


libs/shared/shared/github/__init__.py
```diff
@@ -16,11 +16,14 @@
...
 
 
-def get_pem_overrides():
-    pem_overrides = {}
-    if settings.GITHUB_SENTRY_APP_ID:
-        pem_overrides[settings.GITHUB_SENTRY_APP_ID] = settings.GITHUB_SENTRY_APP_PEM
-    return pem_overrides
+def get_pem_overrides(app_id: str) -> tuple[str, ...] | None:
+    if (
+        settings.GITHUB_SENTRY_APP_ID is not None
```
ajay-sentry, line 21: can you just do settings.GITHUB_SENTRY_APP_ID and .... rather than the is not None

libs/shared/shared/github/__init__.py
```diff
@@ -39,13 +42,19 @@ def load_pem_from_path(pem_path: str) -> str:
     raise Exception("Unknown schema to load PEM")
 
 
-def get_pem(pem_name: str | None = None, pem_path: str | None = None) -> str:
+def get_pem(app_id: str, service: str, pem_path: str | None = None) -> str:
+    # override pem path if provided
+    # otherwise check if the app id is in overrides
+    # otherwise use the default pem name
```
ajay-sentry, line 48: this should say pem name from service rather than default right

libs/shared/shared/github/__init__.py
joseph-sentry: i think it's better to explicit, since we don't want to reject any falsy values, just None

ajay-sentry (approved):
aight

PR #504, opened 2025-09-19
fix: update feedback form in TA PR comment

joseph-sentry:
we need to update the link because there's a new form :)


ajay-sentry (approved):
good

PR #512, opened 2025-09-25
fix: move tasks to experimental module and undo notifier tauploadcontext

joseph-sentry:
- clarify the tasks that are simply experimental and not actually meant for use in prod currently
- remove TAUploadContext type from commit in BaseNotifier constructor type signature: we don't want to have to worry about this case since it's not being used in prod

### DIFF
--- a/apps/worker/tasks/experimental/test_analytics_notifier.py
+++ b/apps/worker/tasks/experimental/test_analytics_notifier.py
@@ -55,6 +55,12 @@ class TestAnalyticsNotifierTask(
     """
     Send test analytics notifications while ensuring compliance with Sentry's
     data retention policies.
+
+    ---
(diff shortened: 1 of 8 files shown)


apps/worker/tasks/experimental/test_analytics_notifier.py
```diff
@@ -282,7 +288,7 @@ def notification_preparation(
 
         notifier = TestResultsNotifier(
             repo,
-            upload_context,
+            upload_context,  # TODO: this won't work because test results notifier doesn't currently support the TAUpload Context
```
ajay-sentry, line 291: can you elaborate on "this won't work"

Should we just comment out the whole param to avoid the footgun?

PR #517, opened 2025-09-30
fix: don't override existing paid plans when account linking

giovanni-guidini:
This was recently surfaced as an error in support: a customer had a paid plan via codecov, then installed the Sentry app. The account linking then created a new `Account` for this user in the sentry merge plan, thus overriding their existing plan.

We don't want this to happen. At this time we will just block account linking. Later on, when you can actually purchase the sentry merge plan we will effectively cancel your existing codecov plan in favor of managing payments in Sentry.

### Legal Boilerplate

### DIFF
--- a/apps/codecov-api/api/sentry/tests/test_views.py
+++ b/apps/codecov-api/api/sentry/tests/test_views.py
@@ -550,6 +550,18 @@ class AccountLinkViewTests(TestCase):
             PlanName.ENTERPRISE_CLOUD_YEARLY.value, TierName.ENTERPRISE.value, True
         )
 
+    def test_account_link_blocks_sentry_monthly_plan(self):
+        """Test that account linking is blocked for Sentry Monthly plan"""
+        self._test_account_link_blocks_paid_plan(
+            PlanName.SENTRY_MONTHLY.value, TierName.SENTRY.value, True
+        )
+
+    def test_account_link_blocks_sentry_yearly_plan(self):
+        """Test that account linking is blocked for Sentry Yearly plan"""
+        self._test_account_link_blocks_paid_plan(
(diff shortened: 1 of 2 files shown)


apps/codecov-api/api/sentry/views.py
```diff
@@ -89,16 +94,70 @@ def account_link(request, *args, **kwargs):
...
+            # Check if the organization has an existing paid plan that should
+            # be preserved
+            # FIXME: This is temporary (famous last words) and will be changed
+            # once you can actually purchase prevent plan in Sentry.
+            # At that point the plan is to _cancel_ existing codecov plans and
+            # just use the Sentry one.
+            if existing_owner.plan:
+                try:
+                    plan_obj = Plan.objects.get(name=existing_owner.plan)
+                    if plan_obj.paid_plan and not plan_obj.is_sentry_plan:
```
ajay-sentry, line 107: we can just use .paid_plan and scrap the is_sentry_plan condition since that's the codecov paid plan for sentry users that gives them 5 seats for free

apps/codecov-api/api/sentry/tests/test_views.py
```diff
@@ -509,6 +514,246 @@ def test_conflict_with_active_account(self):
...
+
+        # Verify account was created
+        account = Account.objects.get(sentry_org_id="123456789")
+        self.assertEqual(account.plan, PlanName.SENTRY_MERGE_PLAN.value)
+
+        # Verify owner was linked to new account
+        existing_owner.refresh_from_db()
+        self.assertEqual(existing_owner.account, account)
+
+    def test_account_link_blocks_legacy_pro_monthly_plan(self):
```
ajay-sentry, line 659: reminder to scrap these tests too

PR #518, opened 2025-09-30
support downloading uploads from legacy in the UI

giovanni-guidini:
The issue here is that legacy uploads are processed via shelter, but don't use the `::::` separator in the repo slug. See [here](https://github.com/codecov/shelter/blob/90757e869ee19e5f04e64cc5221ff5ca422bfb3c/upload-ingress/main.py#L751) that

```python
repo_slug = f"{repo.owner_username}/{repo.repo_name}"
storage_path=f"shelter/v4/{service}/{repo_slug}/{commit_sha}/{reportid}.txt"
```

and this breaks on the validation code that is being changed. I don't know how long this has been happening, but many months in the least.

These changes support these paths that start with `shelter/v4` (from legacy uploads).

### Legal Boilerplate


apps/codecov-api/upload/tests/test_upload_download.py
```diff
@@ -147,3 +147,53 @@ def test_invalid_shelter_path(self):
             data={"path": "shelter/github/codecovtest::::some-other-repo"},
         )
         assert response.status_code == 404
+
+    @patch("shared.storage.MinioStorageService.create_presigned_get")
+    def test_valid_shelter_path_with_colons(self, create_presigned_get):
```
ajay-sentry, line 152: this should already be tested above right?

giovanni-guidini: I actually didn't find on when skimming through, so decided to just leave it there

PR #519, opened 2025-09-30
Max/upload retry

maxweng-sentry:
**Summary**
This PR enables admin users to rerun failed uploads directly from the breadcrumbs UI.

**Motivation**
Previously, failed uploads required manual intervention to reprocess. This feature allows admins to requeue and rerun these uploads more easily from the breadcrumb menu.

**Implementation Details**

- Added an action section to each breadcrumb.
- For breadcrumbs associated with failed processing or compiling upload milestones, a ‘Rerun Upload’ button is displayed.
- Clicking the button requeues the corresponding uploads in Redis and triggers the task responsible for compilation and processing.
- Access to this action is restricted to admin users.

**Testing**

- Verified reruns correctly requeue uploads and complete successfully.
- Confirmed the button only appears for failed upload states.
- Created tests to confirm various error and checks are working.

### DIFF
--- a/libs/shared/shared/django_apps/upload_breadcrumbs/admin.py
+++ b/libs/shared/shared/django_apps/upload_breadcrumbs/admin.py
@@ -607,6 +607,8 @@ class UploadBreadcrumbAdmin(admin.ModelAdmin):
         log.info(
             f"Data: {data}, Upload IDs: {obj.upload_ids}, Milestone: {data.get('milestone')}"
         )
+
+        # Only consider uploads that are in the PROCESSING_UPLOAD or COMPILING_UPLOADS milestone since those are the ones that the upload task handles
         if (
             data.get("milestone")
             in [Milestones.PROCESSING_UPLOAD.value, Milestones.COMPILING_UPLOADS.value]


libs/shared/shared/django_apps/upload_breadcrumbs/admin.py
```diff
@@ -519,3 +547,340 @@ def has_add_permission(self, request: HttpRequest, obj: Any = None) -> bool:
...
     def has_change_permission(self, request: HttpRequest, obj: Any = None) -> bool:
         return False
+
+    def get_urls(self):
+        """Add custom URLs for the resend upload functionality."""
+        urls = super().get_urls()
+        custom_urls = [
+            path(
+                "<path:object_id>/resend-upload/",
+                self.admin_site.admin_view(self.resend_upload_view),
```
ajay-sentry, line 557: is there a world where we can move this to the urls file? I'm wondering if there's a world where we want to call this endpoint outside of admin in the future, or at least make refactoring that easier later on

https://github.com/codecov/umbrella/blob/4c645681c6cd69b67336a56f35b9f574219c38d8/apps/codecov-api/codecov/urls.py#L28

libs/shared/shared/django_apps/upload_breadcrumbs/admin.py
```diff
@@ -1,29 +1,46 @@
...
     Endpoints,
     Errors,
     Milestones,
     UploadBreadcrumb,
 )
 from shared.django_apps.utils.paginator import EstimatedCountPaginator
+from shared.helpers.redis import get_redis_connection
+
+# Import TaskService conditionally - it may not be available in all contexts
+try:
```
ajay-sentry, line 33: Is this because TaskService is a worker only concept?

maxweng-sentry: Yeah I think so. It causes an error when running `make devenv` without the conditional import

libs/shared/shared/django_apps/upload_breadcrumbs/admin.py
```diff
@@ -519,3 +547,340 @@ def has_add_permission(self, request: HttpRequest, obj: Any = None) -> bool:
...
+            log.error(
+                "Commit not found in database - cannot resend upload",
+                extra={
+                    "repo_id": breadcrumb.repo_id,
+                    "breadcrumb_id": breadcrumb.id,
+                },
+            )
+            return False
+
+        # Test Redis connection first
```
ajay-sentry, line 711: I don't think we'll need this block

PR #543, opened 2025-10-30
feat: add composite index on repo id, outcome and timestamp for the TA DB

adrianviquez:
Adding another index as to expedite API calls when outcome is specified in the filtering

---

> [!NOTE]
> Adds a composite DB index on `ta_timeseries.testrun(repo_id, outcome, timestamp)` (and syncs `repo_id` index) and updates evals tests to use approximate float assertions.
> 
> - **Database (ta_timeseries)**:
>   - Add composite index `ta_ts_repo_id_outcome_time_idx` on `testrun(repo_id, outcome, timestamp)` via migration `0037` (with TimescaleDB options).
>   - Sync model state for existing `repo_id` index `ta_ts__repo_id_idx`.
>   - Reflect both indexes in `Testrun` model `Meta.indexes`.
> - **Tests**:
>   - Use `pytest.approx` for floating-point assertions in `api/public/v2/tests/test_api_evals_viewset.py` to avoid precision issues.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c0f13d48ad1c29990b0979e886e7f254d778a869. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/codecov-api/api/public/v2/tests/test_api_evals_viewset.py
+++ b/apps/codecov-api/api/public/v2/tests/test_api_evals_viewset.py
@@ -192,11 +192,17 @@ class EvalsViewSetTestCase(APITestCase):
         assert data["totalItems"] == 3
         assert data["passedItems"] == 2
         assert data["failedItems"] == 1
-        assert data["avgDurationSeconds"] == (10.0 + 20.0 + 30.0) / 3
-        assert data["avgCost"] == (5.0 + 7.0 + 3.0) / 3
+        assert data["avgDurationSeconds"] == pytest.approx((10.0 + 20.0 + 30.0) / 3)
+        assert data["avgCost"] == pytest.approx((5.0 + 7.0 + 3.0) / 3)
         assert data["scores"] == {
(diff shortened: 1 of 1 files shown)


libs/shared/shared/django_apps/ta_timeseries/models.py
```diff
@@ -89,6 +89,14 @@ class Meta:
                 name="ta_ts__repo_branch_time_idx",
                 fields=["repo_id", "branch", "timestamp"],
             ),
+            models.Index(
+                name="ta_ts__repo_id_idx",
+                fields=["repo_id"],
+            ),
```
calvin-codecov: should we have a migration using `migrations.SeparateDatabaseAndState` to add the index to django state first like migration 0032 follows up for 0030? I think if we add these indexes here without adding to django state, it may autogenerate a redundant migration that would try to add to both SQL and django state even though the SQL index already exists. Same for the new index that is being added in this.

PR #558, opened 2025-11-11
check uploads for notify

maxweng-sentry:
https://linear.app/getsentry/issue/CCMRG-1821/we-should-not-let-the-notify-task-run-if-there-are-still-pending

Checks to see if there are any uploads still pending before triggering the notify task. Also cleans up any stalled uploads so that future uploads can successfully finish.

### Legal Boilerplate

### DIFF
--- a/apps/worker/tasks/tests/unit/test_upload_finisher_task.py
+++ b/apps/worker/tasks/tests/unit/test_upload_finisher_task.py
@@ -181,6 +181,7 @@ class TestUploadFinisherTask:
             branch="thisbranch",
             ci_passed=True,
             repository__branch="thisbranch",
+            repository__updatestamp=None,
             repository__author__unencrypted_oauth_token="testulk3d54rlhxkjyzomq2wh8b7np47xabcrkx8",
             repository__author__username="ThiagoCodecov",
             repository__author__service="github",
@@ -208,6 +209,11 @@ class TestUploadFinisherTask:
         assert result == {"notifications_called": True}
         dbsession.refresh(commit)
         assert commit.message == "dsidsahdsahdsa"
+
+        # Verify repository timestamp is updated
+        dbsession.refresh(commit.repository)
+        assert commit.repository.updatestamp is not None
+        assert (datetime.now(tz=UTC) - commit.repository.updatestamp).seconds < 60
 
         mock_checkpoint_submit.assert_any_call(
             "batch_processing_duration",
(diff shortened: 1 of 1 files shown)


apps/worker/tasks/upload_finisher.py
```diff
@@ -347,6 +347,12 @@ def run_impl(
 
         except SoftTimeLimitExceeded:
             log.warning("run_impl: soft time limit exceeded")
+            # Clean up orphaned state so future finishers can proceed
+            state.mark_uploads_as_merged(upload_ids)
```
thomasrockhu-codecov: I think this is wrong, this will mark all uploads as `merged` correct? But if we hit a `SoftTimeLimitExceeded`, I'm guessing this is not true

apps/worker/tasks/manual_trigger.py
```diff
@@ -96,6 +97,26 @@ def process_impl_within_lock(
...
             if not upload.state or upload.state_id == UploadState.UPLOADED.db_id:
                 still_processing += 1
+
+        # Also check Redis processing state to prevent race conditions
+        # where uploads have been marked complete in DB but are still
+        # being processed/merged by the finisher task
+        state = ProcessingState(repoid, commitid)
+        upload_numbers = state.get_upload_numbers()
+
+        if upload_numbers.processing > 0 or upload_numbers.processed > 0:
```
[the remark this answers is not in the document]
thomasrockhu-codecov: yeah, this is true, but there's another state `merged`.

apps/worker/tasks/tests/unit/test_upload_finisher_task.py
```diff
@@ -1156,3 +1255,49 @@ def test_reconstruct_processing_results_returns_empty_when_no_uploads_found(
 
         # Verify empty list returned when no uploads found
         assert result == []
+
+    @pytest.mark.django_db
+    def test_upload_finisher_updates_repository_timestamp(
```
thomasrockhu-codecov: Does this test need to exist as a standalone?

apps/worker/tasks/manual_trigger.py
```diff
@@ -96,6 +97,26 @@ def process_impl_within_lock(
...
+            log.info(
+                "Retrying ManualTriggerTask. Redis shows uploads still being processed.",
+                extra={
+                    "repoid": repoid,
```
thomasrockhu-codecov: nit: alphabetize

apps/worker/tasks/manual_trigger.py
```diff
@@ -96,6 +97,26 @@ def process_impl_within_lock(
         for upload in uploads:
             if not upload.state or upload.state_id == UploadState.UPLOADED.db_id:
                 still_processing += 1
+
+        # Also check Redis processing state to prevent race conditions
```
thomasrockhu-codecov: I'm not sure I buy this, I think it makes more sense to use either the DB or Redis as the source of truth even with a race condition.

thomasrockhu-codecov (changes requested):

armenzg:
@sentry review

PR #561, opened 2025-11-13
CCMRG-1864 Add Dead Letter Queue (DLQ) for Failed Tasks

drazisil-codecov:
# Add Dead Letter Queue (DLQ) for Failed Tasks

## Summary
Implements a Dead Letter Queue (DLQ) system to prevent data loss when Celery tasks exhaust all retries. Tasks are saved to Redis for manual inspection and recovery. Recovered tasks behave exactly like normal retries.

## Problem
When tasks exhaust all retries, they are silently dropped, leading to data loss and no recovery mechanism.

## Solution

**DLQ Configuration** (`libs/shared/shared/celery_config.py`)
- `DLQ_TTL_SECONDS`: Configurable TTL (default: 7 days)
- `DLQ_KEY_PREFIX`: Redis key prefix (`task_dlq`)

**Base Task Support** (`apps/worker/tasks/base.py`)
- `_save_to_task_dlq()`: Generic method for all tasks to save to DLQ
- Handles `UserYaml` serialization and ensures JSON compatibility
- Metric: `worker_task_counts_dlq_saved_total`

**DLQ Recovery Task** (`apps/worker/tasks/dlq_recovery.py`)
- Actions: `list`, `recover`, `delete`
- Automatically restores upload arguments to Redis for `UploadTask` recovery

**Django Admin Interface** (`apps/codecov-api/core/admin_dlq.py`)
- Web-based DLQ management at `/admin/dlq/`

**Upload Chain Integration**
- **UploadTask**: Saves when max retries exceeded, captures upload arguments via `get_all_arguments()`
- **UploadProcessorTask**: Saves when max retries exceeded during processing
- **UploadFinisherTask**: Saves when max retries exceeded acquiring report lock

**Key Feature: Upload Argument Preservation**
- `get_all_arguments()` reads arguments from Redis without consuming them
- Arguments saved to DLQ and automatically restored during recovery
- Ensures recovered tasks schedule downstream tasks correctly

## Usage

**Django Admin**: `/admin/dlq/` to list, recover, or delete entries

**Programmatic**:
```python
from services.task.task import celery_app
from shared.celery_config import dlq_recovery_task_name

celery_app.send_task(dlq_recovery_task_name, kwargs={
    "action": "recover",
    "dlq_key": "task_dlq/app.tasks.upload.Upload/123/abc/coverage"
}).get()
```

## Configuration
```yaml
setup:
  tasks:
    celery:
      dlq_ttl_seconds: 604800  # 7 days (default)
```

## Testing
Unit tests cover DLQ save, recovery, upload argument restoration, and all three upload chain tasks. All tests passing.

### DIFF
--- a/apps/codecov-api/core/templates/admin/core/dlq_list.html
+++ b/apps/codecov-api/core/templates/admin/core/dlq_list.html
@@ -1,52 +1,51 @@
 {% extends "admin/base_site.html" %}
-{% load i18n admin_urls static admin_list %}
-
-{% block title %}Dead Letter Queue{% endblock %}
+{% load i18n admin_urls %}
 
 {% block content %}
-<h1>Dead Letter Queue</h1>
+<div class="module">
+  <h1>Dead Letter Queue</h1>
 
-{% if task_name_filter %}
-<p>Filtered by task name: <strong>{{ task_name_filter }}</strong></p>
-{% endif %}
-
-<p>Total DLQ keys: <strong>{{ total_keys }}</strong></p>
-
-{% if keys %}
-<table>
-  <thead>
-    <tr>
-      <th>DLQ Key</th>
-      <th>Task Count</th>
-      <th>TTL (seconds)</th>
-      <th>Actions</th>
-    </tr>
-  </thead>
-  <tbody>
-    {% for key_info in keys %}
-    <tr>
-      <td><code>{{ key_info.key }}</code></td>
-      <td>{{ key_info.count }}</td>
-      <td>{{ key_info.ttl_seconds }}</td>
-      <td>
-        <a href="{% url 'admin:core_dlq_recover' dlq_key=key_info.key %}" class="button">Recover</a>
-        <a href="{% url 'admin:core_dlq_delete' dlq_key=key_info.key %}" class="button" onclick="return confirm('Are you sure you want to delete this DLQ key?');">Delete</a>
-      </td>
-    </tr>
-    {% endfor %}
-  </tbody>
-</table>
-{% else %}
-<p>No DLQ keys found.</p>
-{% endif %}
-
-<form method="get" action="{% url 'admin:core_dlq_list' %}">
-  <label for="task_name">Filter by task name:</label>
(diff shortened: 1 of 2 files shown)


apps/codecov-api/core/templates/admin/core/dlq_list.html
```diff
@@ -0,0 +1,52 @@
+{% extends "admin/base_site.html" %}
```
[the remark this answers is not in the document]
drazisil-codecov: There is no db model for the dead letter queue, so it needs a little bit of custom template. Will attempt to reformat it more in line with other pages .

apps/worker/tasks/upload.py
```diff
@@ -175,6 +175,22 @@ def arguments_list(self):
             for arg in arguments:
                 yield orjson.loads(arg)
 
+    def get_all_arguments(self) -> list[dict]:
```
thomasrockhu-codecov: this function can be combined with the one above

apps/worker/tasks/dlq_recovery.py
```diff
@@ -0,0 +1,241 @@
...
+            return self._list_dlq_keys(redis_conn, task_name_filter)
+        elif action == "recover":
+            if not dlq_key:
+                return {
+                    "success": False,
+                    "error": "dlq_key required for recover action",
+                }
+            return self._recover_tasks(redis_conn, dlq_key)
+        elif action == "delete":
+            if not dlq_key:
```
thomasrockhu-codecov: I think all of these CRUD operations don't make sense per se, I think it's worth considering making dead letter tasks into an object to avoid this work.

thomasrockhu-codecov (changes requested):
Major points
1. The admin view doesn't make sense here, it would be better to be based off of a model, please consider making this a table
2. The work that's done in the specific tasks (e.g. `upload`, `upload_finisher`, `upload_processor`) should be encapsulated in the base task. 
3. The admin view should not be the priority, you can split it out into another PR

This feels completely vibe-coded, and as a result, there are a lot of inefficiencies. Before opening a PR, please take the time to go through the logic and consider if the architecture is correct. If you are unsure, talk it through with another engineer.

PR #563, opened 2025-11-15
feat: add post/get endpoints to create/fetch ta export data

adrianviquez:
We're creating 2 endpoints for posting and fetching TA data export.

The post route will create a worker task to actually do the logic to generate TA data to a gcp bucket, provided and passed as part of the params.

The get route will be used to get the status of a specific task's state, which will be polled by the code initiating the data export.

Tests were written for both new endpoints.

As well, the previous endpoint containing the data creation process was deleted, and that logic will be moved to the worker code in a followup PR

---

> [!NOTE]
> Add POST/GET endpoints to schedule and query Test Analytics exports via Celery, wire up new export task routing, and remove the legacy EU endpoint with corresponding tests.
> 
> - **API (Sentry)**:
>   - Add `POST /internal/test-analytics/exports/` (`create_ta_export`) to schedule TA exports via `TaskService` to `export_test_analytics_data_task_name` with required params (`integration_names`, `gcp_project_id`, `destination_bucket`, `destination_prefix`).
>   - Add `GET /internal/test-analytics/exports/<task_id>/` (`get_ta_export`) to fetch task status/results using `AsyncResult`.
>   - Remove legacy `POST /internal/test-analytics/eu/` endpoint.
> - **Celery Config**:
>   - Introduce `TaskConfigGroup.export_test_analytics` and `export_test_analytics_data_task_name`.
>   - Route new task in `BaseCeleryConfig.task_routes`.
> - **Tests**:
>   - Add comprehensive tests for create/get endpoints and task scheduling/result handling.
>   - Update celery config unit test to include new task route; delete old endpoint tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c9f62f280c61c59d70cf163f1f30bf66fcec33e6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


apps/codecov-api/api/sentry/views.py
```diff
@@ -250,67 +248,211 @@ def account_unlink(request, *args, **kwargs):
...
+        except Exception as e:
+            log.error(
+                "Failed to schedule TA export task",
+                extra={
+                    "integration_name": integration_name,
+                    "error": str(e),
+                },
+                exc_info=True,
+            )
+            task_results.append(
```
ajay-sentry: jw the design choice to also append and send these to the client. Since they're captured in the log anyway, is there additional value we're getting through this?

I'm wondering if we should also fire a sentry exception here too

apps/codecov-api/api/sentry/views.py
```diff
@@ -250,67 +248,211 @@ def account_unlink(request, *args, **kwargs):
...
-            test_runs_per_integration[name] = {}
-            continue
+    successful_tasks = [task for task in task_results if "task_id" in task]
+    log.info(
+        "Completed data export scheduling for test analytics",
+        extra={
+            "total_requested": len(integration_names),
+            "successfully_scheduled": len(successful_tasks),
+            "failed_to_schedule": len(integration_names) - len(successful_tasks),
+            "task_ids": [task.get("task_id") for task in successful_tasks],
```
ajay-sentry: if we do the above , then we can just return the task_ids in a string to the client and simplify a lot of this stuff below

apps/codecov-api/api/sentry/views.py
```diff
@@ -250,67 +248,211 @@ def account_unlink(request, *args, **kwargs):
...
+
+    Returns:
         {
-            "test_runs_per_integration": test_runs_per_integration,
+            "task_id": "celery-task-id-here",
+            "status": "SUCCESS|PENDING|FAILURE|RETRY|STARTED",
+            "result": {...}  # Only present if status is SUCCESS
         }
+    """
+    log.info(
```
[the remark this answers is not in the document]
ajay-sentry: Reason why I suggest that is a lot of these logs just kinda muddy up the logic to me, when really we're just reporting if the result exists / failed/ or is in progress still

apps/codecov-api/api/sentry/views.py
adrianviquez: Wdym, like not logging the total/success/fail breakdown?

We do get value in reporting the status of the job (failed, queued, success, etc) as it will react based on it. For the logs, do you find them not as helpful? These are moreso for our own debugging/logging, which I don't think are super complex and help understand the flow of things, plus often helpful for supporting. Though open to see your thoughts here too

apps/codecov-api/api/sentry/views.py
```diff
@@ -250,67 +248,187 @@ def account_unlink(request, *args, **kwargs):
     )
 
 
-class SentryTestAnalyticsEuSerializer(serializers.Serializer):
-    """Serializer for test analytics EU endpoint"""
+class CreateTestAnalyticsExportSerializer(serializers.Serializer):
```
ajay-sentry: nit: generate might be a clearer verb than create maybe 🤔

apps/codecov-api/api/sentry/views.py
```diff
@@ -250,67 +248,211 @@
...
-        # Fetch all test runs for all repositories in a single query
-        test_runs = Testrun.objects.filter(repo_id__in=repo_id_to_name.keys()).order_by(
-            "repo_id", "-timestamp"
-        )
+    return Response(
+        {
+            "tasks": task_results,
+            "total_tasks": len(task_results),
+            "successfully_scheduled": len(successful_tasks),
+        },
```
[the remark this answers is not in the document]
ajay-sentry: @adrianviquez should def follow this imo, wouldn't want to expose the stack trace to our client

PR #564, opened 2025-11-16
fix: move the error to occur after all retries

thomasrockhu-codecov:
Currently we inflate our FileNotInStorage errors due to retries. This moves the error logic into Sentry at a later time (when retries have been exhausted). This will help better gauge how many instances we actually have.

### Legal Boilerplate

---

> [!NOTE]
> Defers Sentry reporting for `FILE_NOT_IN_STORAGE` until after retry attempts, adding contextual capture in the task and removing early capture in report parsing.
> 
> - **Worker Upload Processing**:
>   - On retryable `FILE_NOT_IN_STORAGE`, schedule a retry; after `MAX_FILE_NOT_FOUND_RETRIES`, capture `FileNotInStorageError` to Sentry with detailed context.
>   - Introduces `MAX_FILE_NOT_FOUND_RETRIES = 1`.
> - **Report Service**:
>   - Stops capturing `FileNotInStorageError` to Sentry when reading raw report; logs and returns a retryable `ProcessingError` instead.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 630cb6dfc69f9903c12d637334560ee20352135b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/worker/tasks/upload_processor.py
+++ b/apps/worker/tasks/upload_processor.py
@@ -1,5 +1,6 @@
 import logging
 
+import sentry_sdk
 from celery.exceptions import SoftTimeLimitExceeded
 from sqlalchemy.orm import Session as DbSession
 
@@ -9,6 +10,7 @@ from services.report import ProcessingError
 from shared.celery_config import upload_processor_task_name
 from shared.config import get_config
 from shared.django_apps.upload_breadcrumbs.models import Errors, Milestones
+from shared.storage.exceptions import FileNotInStorageError
 from shared.upload.constants import UploadErrorCode
 from shared.yaml import UserYaml
 from tasks.base import BaseCodecovTask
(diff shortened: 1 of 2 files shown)


apps/worker/tasks/upload_processor.py
```diff
@@ -117,6 +120,24 @@ def on_processing_error(error: ProcessingError):
                     error=Errors.INTERNAL_RETRYING,
                 )
                 self.retry(max_retries=MAX_RETRIES, countdown=FIRST_RETRY_DELAY)
+            elif (
+                error.is_retryable
+                and self.request.retries >= MAX_FILE_NOT_FOUND_RETRIES
+            ):
+                sentry_sdk.capture_exception(
+                    FileNotInStorageError(
+                        f"File not found in {error.params.get('location')} for upload {arguments.get('upload_id')}"
```
adrianviquez: Thoughts on making the error generic and the parameters dynamic?

If we consider something like "File not found in storage" vs "File not found in <dynamic val> for upload <dynamic val>", the 1st one will show all the logs for this type of error, whereas, _I think_, the 2nd one will show as maany entries in sentry. 

If this behaves how I think, I'd prefer to lower the cardinality of the error (1st approach) and provide the details solely in the context, as you are doing rn

PR #567, opened 2025-11-17
fix: upload task should not block on upload_processing locks

drazisil-codecov:
The prior retry work surfaced that the upload_processing lock was blocking the upload task. This is not desired, we should only block on processing the same upload id.


apps/worker/tasks/upload.py
```diff
@@ -324,56 +308,11 @@ def run_impl(
...
-                    "was_setup": False,
-                    "was_updated": False,
-                    "tasks_were_scheduled": False,
-                    "reason": "too_many_processing_retries",
-                }
+        # Note: We don't check for upload_processing lock here because:
+        # 1. UploadTask should continue processing uploads regardless of processing state
+        # 2. Only UploadFinisherTask blocks on upload_processing lock (via get_report_lock)
+        # 3. UploadProcessorTask handles the upload_processing lock for individual upload processing
+        # The upload_processing lock prevents concurrent report merging, not upload scheduling
```
adrianviquez: Hmm do you have more context on this? You're probably correct about your change, just cautious about moving a piece of code like this

PR #578, opened 2025-11-19
fix: make retriable uploads retriable even if no error

thomasrockhu-codecov:
Makes uploads retriable even if there is no error. This helps us retry the upload if there was another issue that isn't caught by our system.

### Legal Boilerplate

---

> [!NOTE]
> Enables resending uploads from the admin UI regardless of failure status, adds a safeguard when `upload_ids` is `None`, and updates tests accordingly.
> 
> - **Admin (upload_breadcrumbs)**
>   - Always show `resend_upload_button` and enable `resend_upload_action` regardless of failure state.
>   - Remove failure checks in `resend_upload_view`; proceed to resend for any breadcrumb.
>   - Add `_resend_upload` guard: return error if `breadcrumb.upload_ids` is `None`.
> - **Tests**
>   - Update expectations to always render resend button/action.
>   - Change non-failed resend test to expect success instead of error.
>   - Add assertions for success messaging and commit short SHA presence.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3727ba0272a473d1dae11d4903e65303115b9fd1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/libs/shared/shared/django_apps/upload_breadcrumbs/admin.py
+++ b/libs/shared/shared/django_apps/upload_breadcrumbs/admin.py
@@ -691,6 +691,17 @@ class UploadBreadcrumbAdmin(admin.ModelAdmin):
             "resend_timestamp": timezone.now().isoformat(),
         }
 
+        # Check if upload_ids is None
+        if breadcrumb.upload_ids is None:
+            log.error(
+                "Cannot resend upload - breadcrumb has no upload IDs",
+                extra={"breadcrumb_id": breadcrumb.id},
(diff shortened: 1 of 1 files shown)


maxweng-sentry (approved):
I think this looks good. Was wondering if it might be good to allow retried uploads when there aren't upload_ids tied to it though. Maybe like searching the commit from the db?

PR #580, opened 2025-11-21
feat: add max retry metrics to upload flow tasks and notify task

drazisil-codecov:
- Replace retry() calls with safe_retry() in UploadTask, UploadProcessorTask, UploadFinisherTask, and NotifyTask
- Add proper error handling when max retries are exceeded
- Update tests to mock safe_retry() instead of retry()
- All tasks now track worker_task_counts_max_retries_exceeded metric

This ensures consistent retry metric tracking across all upload flow tasks,
matching the implementation from commit 427ec6484b11c38e77e8016e563640a712363717.

### DIFF
--- a/apps/worker/tasks/notify.py
+++ b/apps/worker/tasks/notify.py
@@ -2,7 +2,6 @@ import logging
 
 import sentry_sdk
 from asgiref.sync import async_to_sync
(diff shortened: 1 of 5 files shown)


adrianviquez (commented):
What does safe_retry do/how does it work as opposed to what we have today?

PR #589, opened 2025-12-01
CCMRG-1907 Refactor lock management in tasks to use LockManager

drazisil-codecov:
- Introduced LockManager for consistent lock handling across tasks.
- Updated tasks (ManualTrigger, PreProcessUpload, UploadFinisher, etc.) to utilize LockManager for acquiring locks.
- Added blocking timeout configuration to prevent indefinite waiting on locks.
- Enhanced error handling with LockRetry for better retry logic on lock acquisition failures.
- Removed redundant lock handling code .
- Updated tests to reflect changes in lock management and ensure proper behavior under lock contention scenarios.
- `visibility_timeout` is now adjustable based on if `acks_late` is true.
- Migrated task logging for max retries exceeded to LockManager for central logging.

### DIFF
--- a/apps/worker/tasks/preprocess_upload.py
+++ b/apps/worker/tasks/preprocess_upload.py
@@ -14,6 +14,7 @@ from services.repository import (
     possibly_update_commit_from_provider_info,
 )
 from shared.celery_config import (
+    DEFAULT_LOCK_TIMEOUT_SECONDS,
     PREPROCESS_UPLOAD_MAX_RETRIES,
     pre_process_upload_task_name,
 )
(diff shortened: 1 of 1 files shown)


apps/worker/services/lock_manager.py
```diff
@@ -83,8 +100,24 @@ def locked(self, lock_type: LockType, retry_num=0):
...
-            countdown = min(random.randint(max_retry // 2, max_retry), 60 * 60 * 5)
+            max_retry_cap = 60 * 60 * 5  # 5 hours
+            max_retry_unbounded = 200 * 3**retry_num
+            countdown_unbounded = random.randint(
+                max_retry_unbounded // 2, max_retry_unbounded
+            )
+            countdown = min(countdown_unbounded, max_retry_cap)
+
+            if max_retries is not None and retry_num >= max_retries:
+                log.error(
```
drazisil-codecov: @thomasrockhu-codecov if we hit max retry,  it should hit here. I can't find this log though. So, either we aren't,  or something is wrong.

PR #598, opened 2025-12-04
CCMRG-1914: Fix infinite retry loop bug with visibility timeout re-deliveries

drazisil-codecov:
## Problem

Tasks were getting stuck in infinite retry loops at retry 5 and never exceeding max retries. This happened when:

1. Task reaches retry 5 (`request.retries = 5`)
2. Visibility timeout expires before task completes
3. Redis re-queues the task
4. Celery treats it as a new delivery (not a retry)
5. `request.retries` doesn't increment
6. Task stays at retry 5 and loops forever

Additionally, bundle analysis processor tasks were completing successfully but uploads weren't being marked as complete, causing infinite redispatches. This occurred because:

1. `update_upload()` only called `db_session.flush()` but never `db_session.commit()`
2. If visibility timeout expired before `wrap_up_dbsession()` ran, the transaction was lost
3. Upload state changes were never persisted to the database
4. Uploads remained in "pending" state and kept getting redispatched

## Solution

- **Track total attempts** in task headers (including re-deliveries from visibility timeout expiration)
- **Check both retries and total attempts** before retrying to prevent infinite loops
- **Enhanced LockManager** to properly log errors and send Sentry exceptions when max retries are hit
- **Add explicit `db_session.commit()`** in bundle analysis processor to persist upload state immediately

### Core Fixes
- Added `_get_total_attempts()` method to track total attempts including re-deliveries
- Updated `safe_retry()` to check both retry count and total attempts
- Initialize `total_attempts=1` in task headers when tasks are created
- Increment `total_attempts` in headers when retrying
- Updated `upload_finisher.py` to use new retry logic

### Bundle Analysis Processor Fix
- Added explicit `db_session.commit()` after `update_upload()` to persist upload state
- Added explicit `db_session.commit()` in error handler to persist error state
- Fixed `finally` block to safely handle cases where `result` might not be defined

### LockManager Enhancements
- Added Sentry exception reporting when max retries exceeded
- Enhanced error logging with full context (repoid, commitid, lock_name, report_type, etc.)
- Added proper error tags for filtering in Sentry

### Tests
- Fixed all failing tests to account for new `total_attempts` header
- Updated test assertions to match new retry logic

## Testing

- All existing tests pass
- New logic prevents infinite loops by tracking total attempts
- LockManager properly reports failures to Sentry
- Bundle analysis processor now persists upload state even if task is re-delivered

Fixes CCMRG-1914

---

> [!NOTE]
> Track total attempts across task re-deliveries, enforce max attempts in retries/locks, persist bundle analysis upload state immediately, and harden finisher/notify flows.
> 
> - **Core worker/task framework (BaseCodecovTask)**:
>   - Track total attempts via `headers['attempts']`; expose `attempts` property; override `retry()` to increment attempts.
>   - Add `safe_retry()` and `_has_exceeded_max_attempts()` to cap retries by attempts; use exponential backoff by default.
>   - Emit queue wait metrics inline; improve DB error analysis; robust session wrap-up; minor metric text fixes.
> - **Locking (LockManager)**:
>   - New backoff constants and `LOCK_NAME_SEPARATOR`; enriched `LockRetry` (fields, messages).
>   - Compute capped/jittered countdowns; treat retries as attempts; raise/log and send Sentry when max exceeded; richer log context.
> - **Bundle Analysis**:
>   - Processor/Notifier pass `retry_num=self.attempts` and `max_retries`; on lock max-exceeded, stop retrying gracefully.
>   - Commit upload state immediately after `update_upload()`; persist error state on exceptions; safer cleanup when `result` is None.
>   - Fix `upload.upload_type_id` assignment; type hints/casts; improved retry handling for retryable errors.
> - **Upload/Finisher/Preprocess/Manual Trigger/Notify**:
>   - All lock acquisitions use attempts-based retry and max-retry checks.
>   - Upload Finisher: add `max_retries`, use `UPLOAD_PROCESSOR_MAX_RETRIES`; reconstruct `processing_results` from Redis/DB when missing; idempotency skip when uploads already finalized; avoid notifications while coverage uploads still pending; improved breadcrumbs and logging.
> - **Test Analytics Notifier**:
>   - Add fencing-token debounce with lock context that surfaces max-retries-exceeded and retries with countdown.
> - **Config**:
>   - Add `UPLOAD_PROCESSOR_MAX_RETRIES` and `BUNDLE_ANALYSIS_NOTIFY_MAX_RETRIES`; minor schedule/comment clarifications.
> - **Tests**:
>   - Update for attempts-based logic and new behaviors; add coverage for lock backoff/max-retries, BA processor persistence/edge cases, finisher reconstruction/idempotency.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1824b5071f2b6b1111fabe9b08b9188e51481695. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/worker/tasks/upload_finisher.py
+++ b/apps/worker/tasks/upload_finisher.py
@@ -81,6 +81,8 @@ class UploadFinisherTask(BaseCodecovTask, name=upload_finisher_task_name):
         - Invalidating whatever cache is done
     """
 
+    max_retries = UPLOAD_PROCESSOR_MAX_RETRIES
+
     def _find_started_uploads_with_reports(
         self, db_session, commit: Commit
     ) -> set[int]:


apps/worker/tasks/base.py
```diff
@@ -350,6 +420,31 @@ def run(self, *args, **kwargs):
...
+                # This indicates the task was re-queued by Redis due to visibility timeout
+                # expiration (where total_attempts increments but retry_count doesn't)
+                if total_attempts is not None and int(total_attempts) > retry_count + 1:
+                    log.warning(
+                        f"Task {task.name} re-delivered (visibility timeout expired)",
+                        extra={
+                            "task_id": task.request.id,
+                            "retry_count": retry_count,
+                            "total_attempts": int(total_attempts),
+                            "note": "Task was re-queued by Redis due to visibility timeout expiration",
```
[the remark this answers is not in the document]
drazisil-codecov: ✅ **Fixed!** Added try-except block for `int()` conversion in re-delivery detection. Invalid header values now skip re-delivery detection gracefully and log a debug message.

apps/worker/tasks/base.py
```diff
@@ -232,6 +234,67 @@ def _retry(self, countdown=None):
...
+
+        When visibility timeout expires, Redis re-queues the task but Celery treats
+        it as a new delivery (not a retry), so request.retries doesn't increment.
+        We track total attempts in task headers to catch these re-deliveries.
+
+        Returns:
+            Total number of attempts (retries + re-deliveries)
+        """
+        if not hasattr(self, "request"):
+            return 0
```
thomasrockhu-codecov: I feel like this should return `None` instead no?

apps/worker/tasks/base.py
```diff
@@ -232,6 +234,67 @@ def _retry(self, countdown=None):
...
+
+        # Get retry count from Celery
+        retry_count = self.request.retries if hasattr(self.request, "retries") else 0
+
+        # Get total attempts from headers (tracks re-deliveries)
+        headers = self.request.get("headers", {}) or {}
```
thomasrockhu-codecov: does the `or {}` need to be here?

apps/worker/tasks/base.py
```diff
@@ -232,6 +234,67 @@ def _retry(self, countdown=None):
...
+        total_attempts_header = headers.get("total_attempts", None)
```
thomasrockhu-codecov: if we aren't using `headers` anywhere else here, why is this broken into 2 lines?

apps/worker/tasks/base.py
```diff
@@ -232,6 +234,67 @@ def _retry(self, countdown=None):
...
+            except (ValueError, TypeError):
+                # Invalid header value, fallback to retry count
+                log.warning(
+                    "Invalid total_attempts header value",
+                    extra={
+                        "value": total_attempts_header,
+                        "retry_count": retry_count,
+                    },
+                )
+                return retry_count + 1
```
thomasrockhu-codecov: can this be refactored so that we return at the end of the function and add 1

apps/worker/tasks/base.py
```diff
@@ -248,37 +312,62 @@ def safe_retry(self, max_retries=None, countdown=None, exc=None, **kwargs):
...
-                    "current_retries": current_retries,
-                    "max_retries": task_max_retries,
-                },
-            )
-            TASK_MAX_RETRIES_EXCEEDED_COUNTER.labels(task=self.name).inc()
-            return False
+        # Check both retry count and total attempts to catch visibility timeout re-deliveries
+        # Total attempts = retries + 1 (initial attempt), so we compare against max_retries + 1
+        max_total_attempts = (
+            task_max_retries + 1 if task_max_retries is not None else None
```
thomasrockhu-codecov: there is a lot of +1 stuff going on in this PR, is there a way to eliminate it by using better models for the vars?

apps/worker/tasks/base.py
drazisil-codecov: No, because this is an int counter, and I don't think that None can be incremented. I think we want `0` in this case. Unless python hates that.

apps/worker/services/lock_manager.py
```diff
@@ -108,24 +113,51 @@ def locked(self, lock_type: LockType, retry_num=0, max_retries: int | None = Non
...
                         "lock_name": lock_name,
                         "max_retries": max_retries,
+                        "repoid": self.repoid,
+                        "report_type": self.report_type.value,
                         "retry_num": retry_num,
                     },
+                    exc_info=True,
                 )
+                sentry_sdk.capture_exception(
+                    LockMaxRetriesExceeded(error_msg),
```
thomasrockhu-codecov: can this error message be in the class?

apps/worker/tasks/tests/unit/test_upload_finisher_task.py
```diff
@@ -831,6 +831,9 @@ def test_retry_on_report_lock(self, dbsession, mocker, mock_redis, mock_self_app
 
         task = UploadFinisherTask()
         task.request.retries = 0
+        # Set up request.get() and headers for _get_request_headers()
+        task.request.get = lambda key, default=None: {} if key == "headers" else default
+        task.request.headers = {}
```
thomasrockhu-codecov: what is this doing? seems more complicated than it needs to be for test setup

apps/worker/tasks/base.py
```diff
@@ -232,53 +234,111 @@ def _retry(self, countdown=None):
             if TestResultsFlow.has_begun():
                 TestResultsFlow.log(TestResultsFlow.UNCAUGHT_RETRY_EXCEPTION)
 
-    def safe_retry(self, max_retries=None, countdown=None, exc=None, **kwargs):
-        """
-        Safely retry with max retry limit and proper metrics tracking.
+    def _max_total_attempts(self, max_retries: int | None) -> int | None:
+        """Convert max retries to max total attempts (includes initial attempt)."""
+        return max_retries + 1 if max_retries is not None else None
```
thomasrockhu-codecov: I don't think this needs to be a helper function

thomasrockhu-codecov (changes requested):

apps/worker/tasks/base.py
```diff
@@ -212,13 +212,33 @@ def apply_async(self, args=None, kwargs=None, **options):
...
+        request_headers = self._get_request_headers()
+        if request_headers:
+            headers.update(request_headers)
+        headers.update(kwargs.get("headers", {}) or {})
+        headers["attempts"] = self.attempts + 1
+        kwargs["headers"] = headers
+
+        return super().retry(
+            max_retries=max_retries, countdown=countdown, exc=exc, **kwargs
+        )
```
thomasrockhu-codecov: this function and the one below is doing a lot of protection code, can this not be simplified? the only real thing that we're doing is pulling the attempts and incrementing it

apps/worker/tasks/base.py
```diff
@@ -348,14 +422,36 @@ def run(self, *args, **kwargs):
             task = get_current_task()
             if task and task.request:
                 log_context.task_name = task.name
-                log_context.task_id = task.request.id
+                task_id = getattr(task.request, "id", None)
```
thomasrockhu-codecov: are we sure that `task.request.id` can be None?

apps/worker/tasks/base.py
```diff
@@ -314,26 +399,15 @@ def _analyse_error(self, exception: SQLAlchemyError, *args, **kwargs):
...
+            log.debug(
+                "psycopg2 not available, skipping PostgreSQL-specific error analysis"
+            )
         log.exception(
             "An error talking to the database occurred",
             extra={"task_args": args, "task_kwargs": kwargs},
             exc_info=True,
         )
 
-    def _emit_queue_metrics(self):
```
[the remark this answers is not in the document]
drazisil-codecov: Now inlined

apps/worker/tasks/base.py
```diff
@@ -427,8 +523,9 @@ def wrap_up_dbsession(self, db_session):
     def on_retry(self, exc, task_id, args, kwargs, einfo):
         res = super().on_retry(exc, task_id, args, kwargs, einfo)
         self.task_retry_counter.inc()
-        # Track retry count for better observability
-        retry_count = self.request.retries if hasattr(self, "request") else 0
+        retry_count = (
+            getattr(self.request, "retries", 0) if hasattr(self, "request") else 0
```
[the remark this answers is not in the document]
drazisil-codecov: self.request is None until the task is called

Also, this exact check is used in the Celery.Task class itself.

apps/worker/tasks/upload_finisher.py
```diff
@@ -564,7 +583,26 @@ def _handle_finisher_lock(
                 error=Errors.INTERNAL_LOCK_ERROR,
             )
             UploadFlow.log(UploadFlow.FINISHER_LOCK_ERROR)
-            if self.request.retries >= MAX_RETRIES:
+            if self._has_exceeded_max_attempts(UPLOAD_PROCESSOR_MAX_RETRIES):
+                attempts = self.attempts
+                max_attempts = UPLOAD_PROCESSOR_MAX_RETRIES + 1
```
thomasrockhu-codecov: I think these can just be inlined, not separate vars

apps/worker/tasks/base.py
```diff
@@ -314,26 +399,15 @@ def _analyse_error(self, exception: SQLAlchemyError, *args, **kwargs):
                 )
                 return
         except ImportError:
-            pass
+            log.debug(
```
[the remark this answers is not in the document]
drazisil-codecov: It is not. Thank you for celling out my scope creep.

PR #608, opened 2025-12-10
feat(worker): Centralize retry logic in BaseCodecovTask and fix infinite retry bugs

drazisil-codecov:
## Description
Centralizes retry handling in `BaseCodecovTask.retry()` and fixes infinite retry bugs in bundle analysis tasks.

**Changes:**
- Override `BaseCodecovTask.retry()` to handle `max_retries` checks, logging, metrics, and exponential backoff
- Remove `max_retries` parameter from `LockManager.locked()`; delegate retry limits to tasks
- Remove `safe_retry()` and `_retry()` helper methods
- Add contextual retry logging (`repoid`, `commitid`, `report_type`) nested under `context` key
- Fix infinite retry bugs in `BundleAnalysisProcessorTask` and `BundleAnalysisNotifyTask`
- Refactor tests to remove mocks of internal `retry()` logic and test real behavior

**Impact:**
- Prevents infinite retries when `max_retries` is exceeded
- Standardizes retry behavior across all tasks
- Improves observability with structured retry logs
- Simplifies retry logic by removing redundant helpers

---

> [!NOTE]
> Unifies and hardens retry behavior across worker tasks while simplifying lock handling and improving observability.
> 
> - **Centralized retries**: Override `BaseCodecovTask.retry()` to enforce `max_retries`, apply exponential backoff, update `attempts` header, emit metrics, and add structured context in logs; adds default `max_retries` from config and `_emit_queue_metrics()` helper
> - **Lock manager simplification**: Remove `max_retries` from `LockManager.locked()`; it now raises `LockRetry(countdown)` and logs release in `finally`
> - **Task updates**: Migrate `bundle_analysis_processor`, `bundle_analysis_notify`, `upload`, `upload_finisher`, `preprocess_upload`, `manual_trigger`, `test_results_finisher`, and `test_analytics_notifier` to use `self.request.retries` and the new `retry()`; replace custom/safe retry paths and explicit max-retry checks; handle `MaxRetriesExceededError` where needed
> - **Tests**: Refactor to exercise real retry behavior (no internal retry mocks), add coverage for capped backoff, queue-time metrics, and max-retries-exceeded cases to prevent infinite retries
> - **Minor**: Improve DB error analysis handling, tidy logging/messages, and small API/typing cleanups
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e1947a7ad853b42f2a18e2758a2e6080ba2d85cd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/worker/tasks/base.py
+++ b/apps/worker/tasks/base.py
@@ -367,13 +367,8 @@ class BaseCodecovTask(celery_app.Task):
             "attempts": current_attempts + 1,
         }
 
-        # Use the same logic for Celery's max_retries parameter
-        if max_retries_override != _NOT_PROVIDED:
-            celery_max_retries = max_retries_override
-        elif "max_retries" in other_kwargs:
-            celery_max_retries = other_kwargs.pop("max_retries")
-        else:
-            celery_max_retries = self.max_retries
+        # Use task_max_retries for Celery's max_retries parameter (already extracted above)
+        celery_max_retries = task_max_retries
         retry_kwargs_dict = {
             "countdown": countdown,
             "exc": exc,
@@ -417,6 +412,8 @@ class BaseCodecovTask(celery_app.Task):
     def _emit_queue_metrics(self):
         request = getattr(self, "request", None)
         if request is None:
+            return
+        if not hasattr(request, "get"):
             return
         created_timestamp = request.get("created_timestamp", None)
         if created_timestamp:


apps/worker/tasks/base.py
```diff
@@ -219,77 +224,116 @@ def apply_async(self, args=None, kwargs=None, **options):
...
-                    return {"success": False, "reason": "max_retries"}
+            kwargs: Task kwargs to use when retrying (optional, Celery-specific parameter)
+            max_retries: Maximum number of retries allowed (default: task's max_retries).
+                        Pass None explicitly for unlimited retries (matches Celery behavior).
+            **other_kwargs: Additional kwargs to pass to Celery's retry()
         """
-        current_retries = self.request.retries if hasattr(self, "request") else 0
-        task_max_retries = max_retries if max_retries is not None else self.max_retries
-
+        request = getattr(self, "request", None)
```
thomasrockhu-codecov: can this just return out if `self.request` is `None`? The next line is going to freak out trying to call `retries` on `None`

apps/worker/tasks/base.py
```diff
@@ -49,6 +50,9 @@
 
 log = logging.getLogger("worker")
 
+# Sentinel object to distinguish "not provided" from "explicitly None" for max_retries
+_NOT_PROVIDED = object()
```
thomasrockhu-codecov: this is a little bit weird to make a comparison to `object()`. I'd prefer is this were a singleton class or some other primitive

apps/worker/tasks/base.py
```diff
@@ -219,77 +224,116 @@ def apply_async(self, args=None, kwargs=None, **options):
...
         """
-        current_retries = self.request.retries if hasattr(self, "request") else 0
-        task_max_retries = max_retries if max_retries is not None else self.max_retries
-
+        request = getattr(self, "request", None)
+        current_retries = request.retries if request else 0
+        # Resolve max_retries for checking/logging
+        # If max_retries is not provided, use self.max_retries from task class
+        # If max_retries=None is explicitly provided, means unlimited retries (matches Celery behavior)
+        # self.max_retries is inherited from BaseCodecovTask (TASK_MAX_RETRIES_DEFAULT) or overridden by task
```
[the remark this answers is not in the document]
drazisil-codecov: Good point! Reduced excessive inline comments and consolidated the important information. The code is now much more readable while still maintaining clarity on the key logic.

apps/worker/tasks/base.py
```diff
@@ -219,77 +224,116 @@ def apply_async(self, args=None, kwargs=None, **options):
...
+            task_max_retries = self.max_retries
+        else:
+            # Explicitly provided (could be a number or None for unlimited)
+            task_max_retries = max_retries
+
+        request_kwargs = {}
+        if request and hasattr(request, "kwargs"):
+            request_kwargs = request.kwargs if request.kwargs is not None else {}
+        # Note: kwargs parameter is Celery's special parameter for task kwargs on retry
+        retry_kwargs = kwargs if kwargs is not None else {}
```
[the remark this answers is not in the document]
drazisil-codecov: Thanks for the suggestion! However, using a mutable default (`kwargs={}`) is a Python anti-pattern. I kept `kwargs=None` as the default and handle it in the body, which is safer and follows Python best practices. The code now properly handles both `None` and empty dict cases.

apps/worker/tasks/base.py
```diff
@@ -219,77 +224,116 @@ def apply_async(self, args=None, kwargs=None, **options):
...
+        )
+
+        # Pass max_retries to Celery:
+        # - If not provided, use self.max_retries (inherited from BaseCodecovTask or overridden)
+        # - If explicitly provided (including None), pass as-is (None = unlimited in Celery)
+        if max_retries is _NOT_PROVIDED:
+            # Not provided, use task's max_retries (inherited from BaseCodecovTask or overridden)
+            celery_max_retries = self.max_retries
+        else:
+            # Explicitly provided, pass as-is (None means unlimited in Celery, matches superclass behavior)
```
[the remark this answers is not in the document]
drazisil-codecov: Agreed! Removed excessive comments and simplified the code. The important information is now in the docstring, and the implementation is cleaner and easier to read.

apps/worker/tasks/base.py
```diff
@@ -219,77 +224,116 @@ def apply_async(self, args=None, kwargs=None, **options):
...
+        if all_kwargs.get("repoid") is not None:
+            context["repoid"] = all_kwargs.get("repoid")
+        if all_kwargs.get("report_type") is not None:
+            context["report_type"] = all_kwargs.get("report_type")
         if task_max_retries is not None and current_retries >= task_max_retries:
             log.error(
                 f"Task {self.name} exceeded max retries",
                 extra={
-                    "task_name": self.name,
+                    "context": context if context else None,
```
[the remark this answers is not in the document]
drazisil-codecov: This might break the logging pattern of the context dict though, which is why I initially did it this way. We shall see.

apps/worker/tasks/base.py
```diff
@@ -49,6 +50,14 @@
 
 log = logging.getLogger("worker")
 
+
+# Sentinel class to distinguish "not provided" from "explicitly None" for max_retries
+class _NotProvided:
+    """Sentinel class to distinguish 'not provided' from 'explicitly None' for max_retries"""
+
+
+_NOT_PROVIDED = _NotProvided()
```
drazisil-codecov: @thomasrockhu-codecov I don't understand why a named class is not the same as a unnamed class. This is different from how I do singletons in JavaScript. Also, while the comment says "singleton", is that really what we are doing here? The idea is to check for "set vs not-set", because Celery treats None differently from "not set".

drazisil-codecov (commented):
## Overall Assessment

This is a solid PR that centralizes retry logic and fixes important infinite retry bugs. The approach is clean and well-tested. Here are some suggestions for improvement:

## Strengths

✅ Centralizes retry logic in one place  
✅ Fixes infinite retry bugs  
✅ Good test coverage (95.35%)  
✅ Structured logging with context  
✅ Removes redundant helper methods  

## Suggestions for Improvement

### 1. **Missing Test Coverage** (2 lines in `base.py`)

The coverage report indicates 2 lines are missing coverage. Based on the code, these are likely:
- Lines 254-258: The `if request is None:` edge case branch
- Line 262: The `hasattr(request, "kwargs")` check when `request.kwargs` is falsy

**Recommendation**: Add tests for these edge cases to reach 100% coverage.

### 2. **Redundant Check in `request.kwargs` Handling**

```python:apps/worker/tasks/base.py
request_kwargs = (
    request.kwargs if hasattr(request, "kwargs") and request.kwargs else {}
)
```

The `and request.kwargs` check is redundant since `hasattr` only checks existence. If `request.kwargs` exists but is `None`, this would still use `{}`, which is correct, but the check could be simplified.

**Recommendation**: Consider simplifying to:
```python
request_kwargs = getattr(request, "kwargs", None) or {}
```

### 3. **Inconsistent Retry Check in Bundle Analysis**

In `bundle_analysis_processor.py` line 182, there's still a manual `self.request.retries < self.max_retries` check before calling `self.retry()`. This is now redundant since `self.retry()` handles max_retries internally.

**Recommendation**: Remove the manual check to be consistent with the centralized approach:
```python
if result.error and result.error.is_retryable:
    log.warn(...)
    self.retry(countdown=30 * (2**self.request.retries))
```

### 4. **Code Duplication in `retry()` Method**

The `super().retry()` calls at lines 316-330 are nearly identical, differing only by the `kwargs` parameter.

**Recommendation**: Simplify to:
```python
retry_kwargs_dict = {"countdown": countdown, "exc": exc, "max_retries": celery_max_retries}
if kwargs:
    retry_kwargs_dict["kwargs"] = kwargs
retry_kwargs_dict.update(other_kwargs)
return super().retry(**retry_kwargs_dict)
```

### 5. **Documentation Enhancement**

The docstring for `retry()` could be clearer about the sentinel pattern usage.

**Recommendation**: Add a note explaining why `_NOT_PROVIDED` is used:
```python
"""
Override Celery's retry() to automatically check max_retries and track metrics.

The max_retries parameter uses a sentinel pattern (_NOT_PROVIDED) to distinguish:
- Not provided: Uses self.max_retries from the task class
- Explicitly None: Unlimited retries (matches Celery behavior)
- Explicit integer: Override the task's max_retries for this retry
"""
```

### 6. **Potential Edge Case: Empty String Values**

The context extraction checks `is not None`, but doesn't handle empty strings. If `commitid=""`, it would still be included in context.

**Recommendation**: Consider filtering out empty strings:
```python
if all_kwargs.get("commitid") not in (None, ""):
    context["commitid"] = all_kwargs.get("commitid")
```

### 7. **Lock Manager Documentation**

The docstring for `LockManager.locked()` mentions that tasks should handle max_retries via `self.retry()`, but it might be helpful to add an example.

**Recommendation**: Add a brief example in the docstring showing the pattern:
```python
"""
Raises:
    LockRetry: If lock cannot be acquired, with countdown for retry scheduling.
              The calling task should handle max_retries checking via self.retry().
              
Example:
    try:
        with lock_manager.locked(LockType.UPLOAD, retry_num=self.request.retries):
            # do work
    except LockRetry as e:
        self.retry(countdown=e.countdown)
"""
```

## Minor Issues

- Line 184 in `bundle_analysis_processor.py`: `log.warn` should be `log.warning` (though `warn` is deprecated but still works)

## Questions

1. Should the `request is None` case ever happen in production? If not, should it raise an exception instead of silently using defaults?
2. Is there a reason to keep the manual `max_retries` check in bundle_analysis_processor.py, or should it be removed for consistency?

Overall, this is a well-executed refactoring that improves code maintainability and fixes critical bugs. The suggestions above are mostly minor improvements and edge case handling.

apps/worker/services/lock_manager.py
```diff
@@ -88,17 +99,21 @@ def locked(self, lock_type: LockType, retry_num=0, max_retries: int | None = Non
...
+                try:
+                    yield
+                finally:
+                    # Ensure lock release is logged even if an exception occurs (e.g., timeout)
+                    # The Redis lock's __exit__ will release the lock automatically
+                    lock_duration = time.time() - lock_acquired_time
+                    log.info(
+                        "Releasing lock",
+                        extra={
+                            "repoid": self.repoid,
```
thomasrockhu-codecov: nit: alphabetize

apps/worker/tasks/base.py
```diff
@@ -219,77 +224,116 @@ def apply_async(self, args=None, kwargs=None, **options):
...
-        current_retries = self.request.retries if hasattr(self, "request") else 0
-        task_max_retries = max_retries if max_retries is not None else self.max_retries
-
+        request = getattr(self, "request", None)
+        current_retries = request.retries if request else 0
+        # Resolve max_retries for checking/logging
+        # If max_retries is not provided, use self.max_retries from task class
+        # If max_retries=None is explicitly provided, means unlimited retries (matches Celery behavior)
+        # self.max_retries is inherited from BaseCodecovTask (TASK_MAX_RETRIES_DEFAULT) or overridden by task
+        if max_retries is _NOT_PROVIDED:
```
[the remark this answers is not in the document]
thomasrockhu-codecov: thanks AI, can we rename `max_retries` then to something like `max_retries_override`

apps/worker/tasks/base.py
```diff
@@ -219,77 +229,99 @@ def apply_async(self, args=None, kwargs=None, **options):
...
+                request.kwargs if hasattr(request, "kwargs") and request.kwargs else {}
+            )
 
-        Returns:
-            True if retry was scheduled
-            False if max retries exceeded
+        if max_retries is _NOT_PROVIDED:
+            task_max_retries = self.max_retries
+        else:
+            task_max_retries = max_retries
```
thomasrockhu-codecov: maybe just `task_max_retries = self.max_retries if max_retries is NOT_PROVIDED else max_retries` or similar

apps/worker/tasks/base.py
```diff
@@ -219,77 +229,99 @@ def apply_async(self, args=None, kwargs=None, **options):
...
+                "countdown": countdown,
+                "current_retries": current_retries,
+                "exception_type": type(exc).__name__ if exc else None,
+                "max_retries": task_max_retries,
+                "task_id": getattr(request, "id", None) if request else None,
+                "task_name": self.name,
+            },
+        )
+
+        celery_max_retries = (
```
thomasrockhu-codecov: does this need to be different than `task_max_retries`

PR #678, opened 2026-01-29
feat: Pre-download bundle uploads before lock to reduce contention

jason-ford-codecov:
- Download upload files before acquiring commit-level lock
- Reduces lock hold time by ~30-50% per bundle
- Add lock_key_suffix to LockManager for future per-bundle locking
- Add extract_bundle_name_from_file helper for future use

---

> [!NOTE]
> **Medium Risk**
> Changes the bundle analysis processing flow to move storage I/O outside the commit-level lock and alters retry/error timing; mistakes could increase retries or cause transient processing failures under load.
> 
> **Overview**
> Reduces bundle-analysis lock contention by **pre-downloading the upload file before acquiring the commit-level lock**, then passing a `pre_downloaded_path` into `BundleAnalysisReportService.process_upload` for ingestion (no in-lock GCS download).
> 
> Adds `temporary_upload_file` to manage the temp download lifecycle and introduces a **pre-lock retry path** when the download is missing/empty (leaving the upload unmodified until a later attempt), with extra safety checks inside the lock in case the file disappears between steps.
> 
> Updates and expands tests to cover the new pre-download behavior, temp-file cleanup, and the updated retry semantics (including ensuring the upload file isn’t re-read from storage during `process_upload`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9eef2b9976bc9eb97af0274a9be25618899e6323. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


drazisil-codecov (commented):
## AI Code Review

### Summary
This PR reduces lock contention by pre-downloading bundle upload files before acquiring the Redis lock. The implementation is well-designed with proper cleanup handling and error cases.

### Status: APPROVED with minor suggestions

---

### Issues Found

#### Issue 1: Misleading comment (Low Priority)
**File:** `apps/worker/services/bundle_analysis/report.py`
**Function:** `extract_bundle_name_from_file`
**Problem:** Comment states "Stop after reading the first ~100 events" but no limit is implemented. The loop continues until `bundleName` is found or EOF.

**Suggested Fix:**
```python
# Option A: Implement the limit
for count, (prefix, event, value) in enumerate(ijson.parse(f)):
    if prefix == "bundleName":
        return value
    if count > 100:
        break

# Option B: Fix the comment
# Returns early when bundleName is found; returns None if not present
```

**Action Required:** `FIX_COMMENT_OR_IMPLEMENT_LIMIT`

#### Issue 2: Unused code added (Low Priority, Intentional)
**Files:** `apps/worker/services/lock_manager.py`, `apps/worker/services/bundle_analysis/report.py`
**Problem:** `lock_key_suffix` parameter and `extract_bundle_name_from_file` helper are added but not used in this PR.
**Note:** PR description indicates this is intentional groundwork for future per-bundle parallelism.

**Action Required:** `NONE` (defer to author discretion)

---

### Verification Checklist
- [x] Cleanup handling is correct (`finally` blocks properly clean up temp files)
- [x] Race conditions addressed (each worker creates own temp file copy)
- [x] Fallback behavior works when pre-downloaded file is missing
- [x] Error handling in `_pre_download_upload_file` is comprehensive
- [x] Test coverage is adequate

---

### For AI Agents
```json
{
  "review_result": "APPROVED_WITH_SUGGESTIONS",
  "blocking_issues": [],
  "non_blocking_issues": [
    {
      "id": "misleading-comment",
      "severity": "low",
      "file": "apps/worker/services/bundle_analysis/report.py",
      "function": "extract_bundle_name_from_file",
      "action": "Either implement 100-event limit or update comment to reflect actual behavior",
      "lines": "24-25"
    }
  ],
  "approved": true
}
```

drazisil-codecov (approved):

thomasrockhu-codecov (commented):
We might be able to take advantage of creating a class that handles this file creation and removal.

that way, we can use the pattern (please think of a better class name)

```python
class BundleAnalysisFile:
    def __enter__(self):
        # do all the downloading stuff here

    def __exit__(self):
        # do all the temp file deletion here


with BundleAnalysisFile as ba:
    with Lockwhateverdo the lock stuff:
        do stuff
```

see https://stackoverflow.com/questions/3774328/implementing-use-of-with-object-as-f-in-custom-class-in-python

using a contextlib is also a good solution if we don't want to make a class

https://docs.python.org/3/library/contextlib.html

apps/worker/services/bundle_analysis/report.py
```diff
@@ -332,7 +355,8 @@ def process_upload(
                 ),
             )
         finally:
-            os.remove(local_path)
+            if should_cleanup_local and os.path.exists(local_path):
```
thomasrockhu-codecov: and yeah, don't need the `finally` block anymore

.gitignore
```diff
@@ -131,4 +131,7 @@ dmypy.json
 *.pem
 junit.xml
 
-.DS_Store
\ No newline at end of file
+.DS_Store
+
+# Cursor (user/IDE rules - do not commit)
+.cursor/
```
thomasrockhu-codecov: Please pull in latest

thomasrockhu-codecov (changes requested):

drazisil-codecov (commented):
The class was renamed from `temporary_upload_file` to `BundleUploadFile` during the refactor, but the test imports and usages weren't updated. This will cause an `ImportError` on all 6 new tests. Suggested fix for the import below — the 6 usages of `temporary_upload_file(...)` in the test bodies also need to be changed to `BundleUploadFile(...)`.

apps/worker/tasks/tests/unit/test_bundle_analysis_processor_task.py
```diff
@@ -10,8 +14,11 @@
 from shared.bundle_analysis.storage import get_bucket_name
 from shared.celery_config import BUNDLE_ANALYSIS_PROCESSOR_MAX_RETRIES
 from shared.django_apps.bundle_analysis.models import CacheConfig
-from shared.storage.exceptions import PutRequestRateLimitError
-from tasks.bundle_analysis_processor import BundleAnalysisProcessorTask
+from shared.storage.exceptions import FileNotInStorageError, PutRequestRateLimitError
+from tasks.bundle_analysis_processor import (
+    BundleAnalysisProcessorTask,
+    temporary_upload_file,
+)
```
drazisil-codecov, line 20:
```suggestion
from tasks.bundle_analysis_processor import (
    BundleAnalysisProcessorTask,
    BundleUploadFile,
)
```

Also need to update the 6 usages in the test bodies:
`with temporary_upload_file(...)` → `with BundleUploadFile(...)`

apps/worker/services/bundle_analysis/report.py
```diff
@@ -224,15 +222,25 @@ def _attempt_init_from_previous_report(
 
     @sentry_sdk.trace
     def process_upload(
-        self, commit: Commit, upload: Upload, compare_sha: str | None = None
+        self,
+        commit: Commit,
+        upload: Upload,
+        pre_downloaded_path: str | None,
```
thomasrockhu-codecov: making a note that this changes the order of arguments, make sure it doesn't break any other instances

thomasrockhu-codecov: Also, given the way this code is written, we don't handle `pre_downloaded_path` well if it's None. This should just be `str`

apps/worker/tasks/bundle_analysis_processor.py
```diff
@@ -67,6 +73,75 @@ def _set_upload_error_and_commit(
...
+            "repoid": repoid,
+            "commitid": commit.commitid,
+            "upload_id": upload.id_,
+        }
+        try:
+            with open(temp_file_path, "wb") as f:
+                storage_service.read_file(
+                    get_bucket_name(), upload.storage_path, file_obj=f
+                )
+            log.info(
```
thomasrockhu-codecov: probably `debug` level and where it read from and the temp_file_path

thomasrockhu-codecov (changes requested):
You need to either handle the `None` case for `pre_downloaded_file` or remove that as a possibility and ensure the Errors are probably handled

apps/worker/services/bundle_analysis/report.py
jason-ford-codecov: There is only 2 callers, it should be good.

jason-ford-codecov: this should be resolved now.

apps/worker/tasks/bundle_analysis_processor.py
jason-ford-codecov: this is resolved now

apps/worker/services/bundle_analysis/report.py
```diff
@@ -241,20 +250,31 @@ def process_upload(
...
             if upload.storage_path != "":
-                with open(local_path, "wb") as f:
-                    storage_service.read_file(
-                        get_bucket_name(), upload.storage_path, file_obj=f
+                # pre_downloaded_path is None only if the calling code skips
+                # temporary_upload_file; in normal flow it's always a path.
+                # An empty file means the pre-download failed before the lock.
+                if (
+                    not os.path.exists(pre_downloaded_path)
+                    or os.path.getsize(pre_downloaded_path) == 0
```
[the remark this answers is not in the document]
jason-ford-codecov: this should be resolved now

apps/worker/tasks/bundle_analysis_processor.py
```diff
@@ -107,6 +232,7 @@ def process_impl_within_lock(
         commit_yaml: UserYaml,
         params: UploadArguments,
         previous_result: list[dict[str, Any]],
+        pre_downloaded_path: str | None = None,
```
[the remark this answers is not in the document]
thomasrockhu-codecov: this is still an issue, should not expect this to be `None`

```suggestion
        pre_downloaded_path: str,
```

thomasrockhu-codecov (approved):
See note

PR #883, opened 2026-04-21
fix(bitbucket): replace retired CHANGE-3022 API endpoints

drazisil-codecov:
## Summary

Bitbucket retired three endpoints on 2026-04-14 (CHANGE-3022), all returning 410. This PR replaces them with current equivalents and removes one deprecated call entirely.

**Dead endpoints fixed (410 → working):**
- `GET /user/permissions/repositories` → removed (`list_permissions`) and replaced (`get_authenticated`)
- `GET /workspaces/{uuid}/permissions` → `GET /workspaces/{slug}/permissions` — was passing UUID where slug is required, causing 404s in `get_is_admin`

**Changes:**
- `get_is_admin`: use `self.data["owner"]["username"]` (slug) instead of `service_id` (UUID)
- `get_authenticated`: `GET /repositories/{workspace}?role=contributor&q=full_name="..."` — presence in results = write access
- `list_permissions()`: deleted entirely (see below)

**Why `list_permissions()` was dropped, not replaced:**

`list_permissions` called `GET /repositories?role=member` to discover repos where a user had collaborator access outside of workspace membership. That endpoint is itself **deprecated** per the Bitbucket swagger spec — the recommended replacement is the workspace-scoped `GET /repositories/{workspace}`.

Since `list_repos` already calls `GET /repositories/{workspace}` for every workspace slug returned by `list_teams()`, and Bitbucket is retiring the "repo-level collaborator without workspace membership" access model, `list_permissions` was redundant. Dropping it means the sync path uses only non-deprecated workspace-scoped calls.

Swagger spec used for response shape verification: https://dac-static.atlassian.com/cloud/bitbucket/swagger.v3.json

## Test plan

- [x] 71 unit + integration tests pass locally
- [x] Tested end-to-end locally via real Bitbucket OAuth flow with two accounts:
  - `sync_repos` calls only workspace-scoped endpoints (`GET /repositories/{workspace}`)
  - Non-member workspaces are correctly skipped — never queried
  - A repo added to a workspace mid-session was picked up on the next sync
  - `get_is_admin` correctly returns 403 for a user who is a workspace member but not an owner
  - `get_authenticated` correctly denies edit access for a user not in the workspace
  - A user who is a member of a workspace cannot see private repos in that workspace they do not have explicit access to — Bitbucket's API only returns repos the requesting user can access, so workspace membership alone is not sufficient
- [ ] Verify `sync_repos` / `sync_teams` no longer produce 410s in Sentry after deploy

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes how Bitbucket edit/admin permissions are determined and how accessible repos are enumerated, which could impact repo sync and access checks if the new workspace-scoped queries behave differently across accounts.
> 
> **Overview**
> Updates the Bitbucket adapter to stop using retired `GET /user/permissions/repositories` and to rely on **workspace-scoped** endpoints instead.
> 
> `get_is_admin` now queries workspace permissions using the workspace *slug* (`owner.username`) rather than the stored UUID, and `get_authenticated` determines write/admin capability by querying `GET /repositories/{workspace}` with `role=contributor` and a `full_name` filter (while still verifying private repo access via a direct repo fetch).
> 
> Repository listing is simplified to enumerate repos only via workspaces returned by `list_teams`; the old `list_permissions()` helper and its associated integration/unit tests and VCR cassettes are removed/updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2bb0fe688030bfb072cacebb81a55944be6d5e3b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


libs/shared/shared/torngit/bitbucket.py
```diff
@@ -274,7 +274,7 @@ async def delete_webhook(self, hookid, token=None):
 
     async def get_is_admin(self, user, token=None):
         user_uuid = "{" + user["service_id"] + "}"
-        workspace_slug = self.data["owner"]["service_id"]
+        workspace_slug = self.data["owner"]["username"]
```
michelletran-sentry, line 277: Just checking: do we receive webhooks on name changes? I want to make sure that users who rename their org/slug (if that's possible) will have it reflected in Codecov.

libs/shared/shared/torngit/bitbucket.py
```diff
@@ -397,31 +380,9 @@ async def _fetch_page_of_repos(self, client, username, token, page):
...
-            - /repositories/{username}
-        In order to get all the repositories a user is part of, we first need to get all the orgs and repo owners
-        - Orgs/Teams can be obtained using the 'list_teams' method
-        - Usernames of repo owners is a bit tricky since Bitbucket doesnt provide an endpoint for this
-            - Solution:
-                Use the 'list_permissions' method to get all repo permissions and exctract owner's username
-                from the repository 'full_name' attribute
-        Once we have all orgs/teams and owner's usernames we should call "/repositories/{username}" endpoint
-        for each of the orgs/teams and owner's usernames.
-        """
```
michelletran-sentry, line 415: I feel like we're losing some potential useful context with the removal of this docstring. Can we update it to better reflect what it's actually doing?

libs/shared/shared/torngit/bitbucket.py
drazisil-codecov: We do not, that was a preexisting issue as well. Will address in a later PR.

libs/shared/shared/torngit/bitbucket.py
drazisil-codecov: new docstring added.

libs/shared/shared/torngit/bitbucket.py
drazisil-codecov: We have #748 pending for GitHub, will create a task to do same for Bitbucket

michelletran-sentry (approved):

PR #918, opened 2026-05-07
fix(bots): ensure deterministic ordering of GitHub app installations

drazisil-codecov:
## Summary

- `DjangoSQLAlchemyOwnerWrapper.get_github_app_installations` had no `ORDER BY`, so the DB could return installations in any order — making `app[0]` non-deterministic across calls
- Added `order_by("id")` to the Django path in `owner_helper.py`
- Added `order_by="GithubAppInstallation.id"` to the SQLAlchemy relationship in `core.py`

## Test plan

- [ ] New test in `TestGithubAppInstallationsOrdering` creates 3 apps, deletes and recreates one (forcing its ID higher than the others), and asserts the wrapper always returns them sorted by `id`
- [ ] All existing `test_github_apps.py` tests pass

---

> [!NOTE]
> **Low Risk**
> Low risk: adds explicit ordering to ORM queries/relationships so callers no longer depend on database return order; main impact is making GitHub app selection deterministic when lists are iterated/indexed.
> 
> **Overview**
> Ensures **deterministic ordering** of `Owner.github_app_installations` across both ORM implementations.
> 
> Adds `ORDER BY id` for Django (`DjangoSQLAlchemyOwnerWrapper.get_github_app_installations`) and sets `order_by="GithubAppInstallation.id_"` on the SQLAlchemy `Owner.github_app_installations` relationship, plus a unit test that verifies the wrapper returns installations sorted by id even after delete/recreate changes row IDs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 69eea4bffdb0195971756f2e894414f00178b1cc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/shared/shared/orms/owner_helper.py
+++ b/libs/shared/shared/orms/owner_helper.py
@@ -9,7 +9,7 @@ class DjangoSQLAlchemyOwnerWrapper:
     @staticmethod
     def get_github_app_installations(owner: Owner | Any):
         if _is_django_model(owner):
-            return owner.github_app_installations.all()
+            return owner.github_app_installations.order_by("id")
         else:
             return owner.github_app_installations
 
--- a/apps/worker/database/models/core.py
+++ b/apps/worker/database/models/core.py
@@ -149,7 +149,7 @@ class Owner(CodecovBaseModel):
         foreign_keys="GithubAppInstallation.ownerid",
         cascade="all, delete",
         passive_deletes=True,
-        order_by="GithubAppInstallation.id",
+        order_by="GithubAppInstallation.id_",
     )
 
     # TODO: Create association between `User` and `Owner` mirroring `codecov-api`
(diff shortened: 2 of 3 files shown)


libs/shared/shared/orms/owner_helper.py
```diff
@@ -9,7 +9,7 @@ class DjangoSQLAlchemyOwnerWrapper:
     @staticmethod
     def get_github_app_installations(owner: Owner | Any):
         if _is_django_model(owner):
-            return owner.github_app_installations.all()
+            return owner.github_app_installations.order_by("id")
```
thomasrockhu-codecov, line 12: shouldn't this be ordered by now based off of `models/core.py`? my sqlalchemy is quite weak, so just going off of what I know of django orm

thomasrockhu-codecov (approved):
You probably want to port this over to django ORM as well to keep consistency, but that's out of scope for this PR. I'd probably also add a test that changes the usage of the GitHubAppInstallations

libs/shared/shared/orms/owner_helper.py
drazisil-codecov: We use either orm or sqlalchemy, not together. So we are not running order on the same dataset twice, which is what I think your concern is.

PR #1167, opened 2026-06-15
fix(notify): send checks on GitHub merge queue branches with only_pulls

drazisil-codecov:
## Summary

GitHub merge queue commits run on `gh-readonly-queue/<base-branch>/pr-N-<sha>` branches and have no associated PR (GitHub's `/commits/{sha}/pulls` API returns nothing for these refs). When a repo configures `only_pulls: true` or `base: pr`, `can_we_set_this_status` blocked the commit with `not_fit_criteria` — silently skipping all notifications and leaving the merge queue unable to unblock.

**Fix:** detect merge queue branches in `can_we_set_this_status` and exempt them from the `only_pulls` / `base: pr` guard, treating them as PR-equivalent. The `branches` filter continues to match against the raw branch name unchanged — customers who want to explicitly include or exclude merge queue commits from their filter should use `gh-readonly-queue/<base>.*` patterns.

## Changes

- `apps/worker/services/notification/notifiers/status/base.py` — add `_merge_queue_base_branch()` helper + update `can_we_set_this_status`
- `apps/worker/services/notification/notifiers/tests/unit/test_status.py` — 4 new cases: `only_pulls` bypass, raw-branch filter blocks merge queue branch, explicit `gh-readonly-queue/…` filter passes, negative `!gh-readonly-queue.*` filter is honored

## Test plan

- [x] New test `test_can_we_set_this_status_merge_queue` covers all four cases — passes locally
- [x] Existing `test_can_we_set_this_status_no_pull` still passes (no regression to normal `only_pulls` behavior)
- [x] Full `TestBaseStatusNotifier` class: 22/22 green
- [ ] Verify by re-triggering a merge queue run after deploy

## Notes

- `only_pulls: true` with merge queues is not documented on docs.codecov.com — follow-up doc update warranted
- The `or ()` dead code on `checks/base.py:98` (`if comparison.pull is None or ():`) is unrelated; worth a separate cleanup

### DIFF
--- a/apps/worker/services/notification/notifiers/status/base.py
+++ b/apps/worker/services/notification/notifiers/status/base.py
@@ -20,13 +20,7 @@ from shared.torngit.exceptions import TorngitClientError, TorngitError
 log = logging.getLogger(__name__)
 
 # Matches GitHub merge queue branch names: gh-readonly-queue/<base-branch>/pr-<number>-<sha>
-_MERGE_QUEUE_RE = re.compile(r"^gh-readonly-queue/(.+)/pr-\d+-[0-9a-f]+$")
-
-
-def _merge_queue_base_branch(branch: str) -> str | None:
-    """Return the target base branch for a GitHub merge queue branch, or None."""
-    m = _MERGE_QUEUE_RE.match(branch)
-    return m.group(1) if m else None
+_MERGE_QUEUE_RE = re.compile(r"^gh-readonly-queue/.+/pr-\d+-[0-9a-f]+$")
 
 
 class StatusNotifier(AbstractBaseNotifier):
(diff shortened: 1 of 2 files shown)


apps/worker/services/notification/notifiers/status/base.py
```diff
@@ -18,6 +19,15 @@
 
 log = logging.getLogger(__name__)
 
+# Matches GitHub merge queue branch names: gh-readonly-queue/<base-branch>/pr-<number>-<sha>
+_MERGE_QUEUE_RE = re.compile(r"^gh-readonly-queue/(.+)/pr-\d+-[0-9a-f]+$")
+
+
+def _merge_queue_base_branch(branch: str) -> str | None:
```
thomasrockhu-codecov: it seems like we are only using this as a bool, we could probably just do that instead of returning the regex match

calvin-codecov (approved):

PR #1168, opened 2026-06-15
feat(graphql): add hasLinkedSentryLogin to Me type

drazisil-codecov:
## Summary

- Adds `hasLinkedSentryLogin: Boolean!` to the `Me` GraphQL type
- Resolver checks `owner.user.sentry_user.exists()` — confirms the `SentryUser` row actually exists rather than trusting the denormalized `sentry_user_id` text field on `Owner`
- Needed so Gazebo can show a deprecation notice to users who log in via Sentry (deprecation date: July 8, 2026)

## For Calvin

This is the backend half. Gazebo needs to query `me { hasLinkedSentryLogin }` and show a deprecation banner when it's `true` — pointing users to reconnect via GitHub/GitLab/Bitbucket before July 8.

### DIFF
--- a/apps/codecov-api/graphql_api/types/me/me.graphql
+++ b/apps/codecov-api/graphql_api/types/me/me.graphql
@@ -34,12 +34,12 @@ type Me {
     before: String
   ): MyOrganizationConnection!
     @cost(complexity: 15, multipliers: ["first", "last"])
-  isSyncingWithGitProvider: Boolean!
-  trackingMetadata: trackingMetadata! # temporary solution to expose the user metadata
-  privateAccess: Boolean
-  termsAgreement: Boolean
-  supportPin: String
   hasLinkedSentryLogin: Boolean!
+  isSyncingWithGitProvider: Boolean!
+  privateAccess: Boolean
+  supportPin: String
+  termsAgreement: Boolean
+  trackingMetadata: trackingMetadata! # temporary solution to expose the user metadata
 }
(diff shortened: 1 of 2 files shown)


apps/codecov-api/graphql_api/types/me/me.py
```diff
@@ -4,6 +4,7 @@
 from graphql import GraphQLResolveInfo
 
 from codecov_auth.models import Owner, OwnerProfile
+from services.sentry import is_sentry_user
```
thomasrockhu-codecov: nit, don't need to fix in this PR, there is a standard way of ordering imports. This shouldn't be in between codecov_auth. I don't have the PEP on hand, but again, don't have to fix now

apps/codecov-api/graphql_api/types/me/me.graphql
```diff
@@ -39,6 +39,7 @@ type Me {
   privateAccess: Boolean
   termsAgreement: Boolean
   supportPin: String
+  hasLinkedSentryLogin: Boolean!
```
thomasrockhu-codecov: I'm guilty of this too (see support pin), can we alphabetize these? Also nit, doesn't have to be this PR

calvin-codecov (changes requested):
I think using owner.user.sentry_user.exists() would be a better indicator

calvin-codecov (approved):

